PR #2, opened 2023-11-22
feat(tezos): dfns-1777: add taquito/tezos integration

ghost:
Add tezos integration for the SDK

We use Taquito library to do that


drunkirishcoder:
Sorry didn't get a chance to go through this PR. but a couple things I'd change structurally.

1. I don't think Taquito is the official Tezos library. let's call the lib `lib-taquito` instead. in case in the future we have to support another lib (like ethersjs vs viem) (`lib-solana` and `lib-tron` are ok because those are the official sdks provided by the blockchain).
2. put the example into a subfolder, I suggest call it `basic-tx`. it will be easier for adding more examples in the future.

examples/tezos/main.ts
```diff
@@ -0,0 +1,101 @@
...
+  const forgedBytes = await forger.forge(forgeable)
+
+  const signedOp = await Tezos.signer.sign(forgedBytes, new Uint8Array([3]))
+
+  const op = await Tezos.rpc.injectOperation(signedOp.sbytes)
+
+  console.log(`operation injected:\n ${op}`)
+
+
+  // We can easily sign mi
```
drunkirishcoder: is this another example? if so, let's put it in a different project.

examples/tezos/michelson-signing/taquito/main.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const Tezos = new TezosToolkit(client)
+
+  const forger = new LocalForger()
+
+  Tezos.setForgerProvider(forger)
+
+  Tezos.setProvider({ signer: wallet })
+
+  // We can easily sign michelson data
```
ghost: For this example, it might not be that useful for the user. This feature can be useful, for example, if you need to send a lambda to a contract to be executed but want to restrict the number of users who can submit a lambda by verifiying the signer's address. It will concern only a few of our user. Is it worth keeping it?

Torakushi: IMO, it doesn't hurt

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  public get address(): string {
+    // Tezos-bound wallets will have a tezos address
+    return this.metadata.address!
+  }
+
+  // publicKey implements the Taquito Signer interface
+  publicKey(): Promise<string> {
+    return Promise.resolve(this.tezosPubKey)
```
drunkirishcoder: don't need to resolve, `return this.pubKey` will do

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  // publicKeyHash implements the Taquito Signer interface
+  publicKeyHash(): Promise<string> {
+    return Promise.resolve(this.metadata.address!)
+  }
+
+  // secretKey implements the Taquito Signer interface
+  secretKey(): Promise<string | undefined> {
+    throw undefined;
```
drunkirishcoder: either return `undefined`, or throw not implemented error. `throw undefined` seems weird and confusing.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (res.status !== 'Active') {
+      throw new DfnsError(-1, 'wallet not active', { walletId, status: res.status })
+    }
+
+    if (res.network !== 'Tezos' && res.network !== 'TezosGhostnet') {
+      throw new DfnsError(-1, 'wallet is not bound to Tezos or TezosGhostnet', { walletId, network: res.network })
+    }
+
+    // get the public key encoded with the right tezos prefix.
+    const tezosPublicKey = this.getEncodedPublicKey(res.signingKey.publicKey, res.signingKey.scheme)
```
drunkirishcoder: I suggest move all this logic into the constructor and reduce the constructor param from 3 to 2.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  private static getEncodedPublicKey(publicKey: string, scheme: string): string {
+    switch (scheme) {
+      case 'EdDSA':
+        return b58cencode(publicKey, prefix[Prefix.EDPK])
+      case 'ECDSA':
+        return b58cencode(publicKey, prefix[Prefix.SPPK])
+      default:
+        throw new Error(`KeyScheme.${scheme} is not supported for tezos`)
+    }
```
drunkirishcoder: we can assume at this point the key is correct and condense all this down to an one line expression

```
this.pubKey = b58cencode(publicKey, prefix[scheme === 'EdDSA ? Prefix.EDPK : Prefix.SPPK])
```

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    return Promise.resolve(this.metadata.address!)
+  }
+
+  // secretKey implements the Taquito Signer interface
+  secretKey(): Promise<string | undefined> {
+    throw undefined;
+  }
+
+  // sign implements the Taquito Signer interface
+  // For now, we only want to support transaction signing
```
drunkirishcoder: correct me if I'm misremembering, but I think we can support all signing. the logic would be

1. if magic byte is tx watermark, call sign with `kind: Transaction`.
2. otherwise, we can prepend magic byte, blake2b hash, and sign with `kind: Hash`.

maybe in the future if there's demand, we can support more kinds, but for now, this should work right? no need to throw error.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  async sign(op: string, magicByte?: Uint8Array | undefined): Promise<{ bytes: string; sig: string; prefixSig: string; sbytes: string; }> {
+    if (magicByte?.toString() !== TxWatermark.toString()) {
```
drunkirishcoder: are the bytes string safe? should we `toHex` them for comparison?

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+      throw new DfnsError(-1, `magicBytes ${magicByte} not supported`, { walletId: this.metadata.id, network: this.metadata.network })
+    }
+    
+    
+    const res = await this.dfnsClient.wallets.generateSignature({
+      walletId: this.metadata.id,
+      body: { kind: 'Transaction', transaction: op },
```
drunkirishcoder: minor, I don't remember, but for consistency, if `op` is not prefixed with `0x`, we should add the prefix.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    
+    const sigPrefix = this.metadata.signingKey.scheme === 'ECDSA' ?  prefix.spsig : prefix.edsig
+    return {
+      bytes: op,
+      sig: b58cencode(signature, prefix.sig),
+      sbytes: res.signedData!.substring(2),
+      prefixSig: b58cencode(signature, sigPrefix),
+    }
+  }
+}
```
drunkirishcoder: nit: can we end the file with a new line.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    
+    
+    const res = await this.dfnsClient.wallets.generateSignature({
+      walletId: this.metadata.id,
+      body: { kind: 'Transaction', transaction: op },
+    })
+      
+    assertSigned(res)
+
+    const signature = res.signature!.encoded!.substring(2)
```
drunkirishcoder: prefer `.replace(/^0x/, '')` to be extra safe.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    
+    const sigPrefix = this.metadata.signingKey.scheme === 'ECDSA' ?  prefix.spsig : prefix.edsig
+    return {
+      bytes: op,
+      sig: b58cencode(signature, prefix.sig),
+      sbytes: res.signedData!.substring(2),
```
drunkirishcoder: see above.

package.json
```diff
@@ -20,6 +20,10 @@
     "@ethersproject/strings": "5.7.0",
     "@ethersproject/transactions": "5.7.0",
     "@solana/web3.js": "1.87.6",
+    "@stablelib/blake2b": "1.0.1",
+    "@taquito/local-forging": "17.4.0",
```
drunkirishcoder: don't think we need this anymore.

examples/libs/tezos/basic-tx/taquito-injection/.env.example
```diff
@@ -0,0 +1,7 @@
+DFNS_API_URL='https://api.dfns.wtf'
+DFNS_APP_ID='<redacted>'
+DFNS_APP_ORIGIN='http://localhost:3000'
+DFNS_CRED_ID='<redacted>'
+DFNS_PRIVATE_KEY='<redacted>'
+DFNS_AUTH_TOKEN='<redacted>'
+DFNS_WALLET_ID='<redacted>'
```
drunkirishcoder: please copy the .env.example format from one of the other examples to demonstrate what the values looks like.

also maybe rename `DFNS_WALLET_ID` to `TEZOS_WALLET_ID`.

examples/libs/tezos/basic-tx/taquito-injection/README.md
```diff
@@ -0,0 +1,47 @@
+# Tezos with Taquito
+
+See Taquito's [Utils/API Specification](https://tezostaquito.io/) for a complete guide.
```
drunkirishcoder: a quick sentence explain what this example does instead.

examples/libs/tezos/basic-tx/taquito-injection/README.md
```diff
@@ -0,0 +1,47 @@
...
+Copy `.env.example` to a new file `.env` and set the following values,
+
+- `DFNS_API_URL` = `https://api.dfns.ninja`
+- `DFNS_APP_ID` = the `App ID` from above
+- `DFNS_APP_ORIGIN` = `http://localhost:3000`
+- `DFNS_CRED_ID` = the `Signing Key Cred ID` from above
+- `DFNS_PRIVATE_KEY` = the private key from the step 'generate a keypair', the newlines should not be a problem
+- `DFNS_AUTH_TOKEN` = the `authToken` from above, the value should start with `eyJ0...`
+- `DFNS_SENDER_WALLET_ID` = a Dfns [wallet](https://docs.dfns.co/dfns-docs/api-docs/beta-wallets-api-and-nfts/create-wallet) ID
+- `DFNS_RECEIVER_WALLET_ID` = a Dfns [wallet](https://docs.dfns.co/dfns-docs/api-docs/beta-wallets-api-and-nfts/create-wallet) ID
```
drunkirishcoder: update this to match the actual env var `TEZOS_WALLET_ID`.

examples/libs/tezos/basic-tx/taquito-injection/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+**note** _the wallet must have XTZ to pay for fees_
+
+## Explanation
+
+The program runs on Tezos Ghostnet. In order to run the code, you would need a KeyECDSA or a KeyEdDSA wallet that holds some XTZ. The program will send 1 ꜩ from one wallet to a second wallet.
```
drunkirishcoder: update to reflect the new changes. Tezos Ghostnet wallet, not `KeyECDSA`, that's going to confuse people.

examples/libs/tezos/basic-tx/taquito-injection/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+```shell
+> ts-node main.ts
+
```
drunkirishcoder: can we paste the output here from the example? also make sure it's proper markdown.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Signer } from '@taquito/taquito'
+import { GetWalletResponse, GenerateSignatureResponse } from '@dfns/sdk/types/wallets'
+import { TxWatermark } from './constant'
+import { error } from 'console';
+
+export type DfnsWalletOptions = {
+  walletId: string
+  dfnsClient: DfnsApiClient
+  maxRetries?: number
+  retryInterval?: number
```
drunkirishcoder: we can remove these. not needed anymore.

examples/libs/tezos/basic-tx/taquito-injection/main.ts
```diff
@@ -0,0 +1,78 @@
...
+  return DfnsWallet.init({
+    walletId: walletId,
+    dfnsClient,
+    maxRetries: 10,
+  })
+}
+
+async function main() {
+  // Replace with the RPC url/ chainId you want
+  const client = new RpcClient('https://ghostnet.ecadinfra.com', 'main')
```
drunkirishcoder: let's move the url to env var `TEZOS_NODE_URL`, follow the style of the other examples.

examples/libs/tezos/basic-tx/taquito-injection/main.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const forgeable = await Tezos.prepare.toForge(prepared)
+
+  console.log(`Operation prepared: ${JSON.stringify(forgeable, null, 2)}`)
+
+  const forgedBytes = await forger.forge(forgeable)
+
+  const signedOp = await Tezos.signer.sign(forgedBytes, new Uint8Array([3]))
+
+  const op = await Tezos.rpc.injectOperation(signedOp.sbytes)
```
drunkirishcoder: the example looks overly complicated for a basic transfer. this is how internally things work with prepare/forge/sign etc. for the purpose of these examples, we want to showcase our integration is actually making the development easy. looking at the doc, this is not what people needs to do to transfer some tez. if everything is correctly wired up, we should be able to just do something like

```
await Tezos.contract.transfer({ to: 'tz2JZdBTEkAYXp53X4r5xW5dQ845zi9Z28mT', amount: 1 });
```

this is the [page](https://github.com/ecadlabs/taquito/blob/master/docs/making_transfers.md#transfer-from-an-implicit-tz1-address-to-a-tz1-address) I'm looking at. it does a transfer and waits for 1 confirmation. very simple.

examples/libs/tezos/basic-tx/taquito-injection/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "taquito-injection",
```
drunkirishcoder: can we rename to `taquito-basic-tx` and remove the`taquito-injection` folder? see how the other folders are structured for example.

packages/lib-taquito/index.ts
Torakushi: It didn't work, I needed to add the async (didn't know)

packages/lib-taquito/index.ts
Torakushi: Yes, I proposed the hash signature before:

https://github.com/dfns/dfns-sdk-ts/compare/efe2e327c59f4eb4d7bbb99f3665c1bc61fcc232..1fac4c74cc729d4d804440ed72b4241007954687#diff-d812c8834490e40a229b60ade16aaade16dfcb2e75d53790edb386001edef469L106-L117

 but I understood that we wanted to support signTransaction and nothing more. (like in ripple (coming soon)) did I misunderstand?

packages/lib-taquito/index.ts
Torakushi: From what I've seen, it is safe. I can however create a function utility to compare byte per byte if you prefer

EDIT: I did a element per element comparison

packages/lib-taquito/index.ts
Torakushi: I tested with or without 0x. It works. But I can add it

examples/libs/tezos/basic-tx/taquito-injection/README.md
Torakushi: That's what I did here no?

https://github.com/dfns/dfns-sdk-ts/pull/2/files#diff-c54003187e87ed445cc5310474b1db63778b38f307d8fa457becbd428bb0a6a0R41

examples/libs/tezos/basic-tx/taquito-injection/main.ts
Torakushi: I understand what you are saying. I tend to disagree though (well it really depends on what we want to showcase :) )
Using the forger, it is way more flexible and the user can customise any part (as he wants). The most interesting thing is to delay the operation injection (security check,...)

I do agree that it is simple to do though
```
await Tezos.contract.transfer({ to: 'tz2JZdBTEkAYXp53X4r5xW5dQ845zi9Z28mT', amount: 1 });
```

> looking at the doc, this is not what people needs to do to transfer some tez. if everything is correctly wired up, we should be able to just do something like

The method I used was described in the getting started of Taquito:
https://tezostaquito.io/docs/operation_flow/#customizing-the-operation-flow

TL,DR: If we don't want to show how flexible Taquito can be, we can totally use the `Tezos.contract.transfer` instead of `preparation + forge`

package.json
Torakushi: We do if we support hash signing (and depending on your answer for

https://github.com/dfns/dfns-sdk-ts/pull/2#discussion_r1416966861

Unless you spoke about local-forging?

examples/libs/tezos/basic-tx/taquito-injection/main.ts
Torakushi: Let's keep it KISS! And go with
```
await Tezos.contract.transfer({ to: 'tz2JZdBTEkAYXp53X4r5xW5dQ845zi9Z28mT', amount: 1 });

```

packages/lib-taquito/index.ts
drunkirishcoder: sorry all the chain variations gets confusing. ripple's issue if I remember is `ECDSA` needs SHA512 but `EdDSA` doesn't hash at all, so we decided to only sign full transaction and no hash signing because hashing wouldn't apply to `EdDSA`.

for Tezos, we do support hash signing, so we can leverage that to sign any non-transaction based operations. I see you made the change already. 👍

packages/lib-taquito/index.ts
drunkirishcoder: sorry for some reason I got confused that `Uint8Array.toString` would try to convert it to UTF8 string. I don't know why I thought that, I think because it was late and mixed up `Uint8` and `UTF8`. that's why I asked if it was "safe". the original question didn't make sense, but I see you made the array comparison. 👍

package.json
drunkirishcoder: yeah I'm referring to `local-forging` seems like no longer a dep we need.

examples/libs/tezos/basic-tx/taquito-injection/README.md
drunkirishcoder: what I mean is for the header and intro section of the README, make it example specific so they get a summary before digging any further. see https://github.com/dfns/dfns-sdk-ts/blob/m/examples/libs/tron/basic-tx/README.md?plain=1#L1-L3

README.md
```diff
@@ -177,6 +177,7 @@ Integrations with other blockchain platforms to make Dapp development frictionle
 
 * [BitcoinJS](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-bitcoinjs/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/bitcoinjs/)
 * [ethers.js v5](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs5/)/[v6](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs6/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ethersjs)
+* [Ripple xrpl.js](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ripple/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ripple)
```
drunkirishcoder: shouldn't this be Taquito?

packages/lib-taquito/index.ts
drunkirishcoder: we can remove  `maxRetries` and `retryInterval`, no need to have those options anymore.

packages/lib-taquito/index.ts
```diff
@@ -0,0 +1,125 @@
...
+    throw new DfnsError(
+      -1,
+      'cannot complete signing synchronously because this wallet action requires policy approval',
+      res
+    )
+  }
+}
+
+export class DfnsWallet implements Signer {
+  private options: Required<DfnsWalletOptions>
```
drunkirishcoder: don't need this private field anymore actually.

README.md
Torakushi: arf... working in parallel on ripple got me confused... sorry will fix

bin/publish.sh
```diff
@@ -6,6 +6,7 @@ declare -a packages=(
     "@dfns/lib-bitcoinjs"
     "@dfns/lib-ethersjs5"
     "@dfns/lib-ethersjs6"
+    "@dfns/lib-ripple"
```
drunkirishcoder: `lib-taquito`

bin/version_all.sh
```diff
@@ -8,6 +8,7 @@ npm version --no-git-tag-version \
     --workspace @dfns/lib-bitcoinjs \
     --workspace @dfns/lib-ethersjs5 \
     --workspace @dfns/lib-ethersjs6 \
+    --workspace @dfns/lib-ripple \
```
drunkirishcoder: `lib-taquito`

README.md
```diff
@@ -177,6 +177,7 @@ Integrations with other blockchain platforms to make Dapp development frictionle
 
 * [BitcoinJS](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-bitcoinjs/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/bitcoinjs/)
 * [ethers.js v5](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs5/)/[v6](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs6/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ethersjs)
+* [Tezos taquito](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-taquito/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/tezos)
```
drunkirishcoder: can we change the example folder to be `taquito` to be aligned with `lib-taquito`.

packages/lib-taquito/addHexPrefix.ts
```diff
@@ -0,0 +1,3 @@
+export function addHexPrefix(inputString: string): string {
+    return inputString.startsWith('0x') ? inputString : '0x' + inputString;
+}
```
drunkirishcoder: small pet peeve, can we make sure all files end with a new line so we don't get this? :) `addHexPrefix`, `constants` and `uint8ArraysEqual.ts`.

Screenshot 2023-12-11 at 12 54 53 PM

drunkirishcoder (approved):

PR #3, opened 2023-11-23
bitcoinjs support

drunkirishcoder:
bitcoinjs sdk integration with examples.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
+**/.env
 **/dist
 **/node_modules
-**/.env
 /tmp
 .DS_Store
 docs-generated/
--- a/README.md
+++ b/README.md
@@ -175,9 +175,9 @@ BaseAuthApi.createUserRegistration()
 
 Integrations with other blockchain platforms to make Dapp development frictionless. More to come...
 
-* [ethers.js v5](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs5/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ethersjs/v5)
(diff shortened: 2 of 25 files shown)


examples/bitcoinjs/basic-tx/main.ts
```diff
@@ -0,0 +1,109 @@
...
+      id: 'sendrawtransaction',
+      method: 'sendrawtransaction',
+      params: [hex],
+    })
+  ).data.result
+}
+
+// change this to an utxo for your wallet
+const utxo = {
+  txid: '33c863366067b8b04a29b9b14692939951c75af92f7f47936d6deb2ec083d26c',
```
ghost: question, I know that the lib can't do coin selection (UTXO retriever)

Should we add in this example how the user can retrieve the utxo of a given address? maybe something like

https://github.com/bitpay/bitcore-explorers/blob/master/lib/insight.js#L81 ?

(I didn't check further, I am just wondering how advanced the user will be)

drunkirishcoder: it's definitely value added, but I think it's out of the scope of what our examples should show. Let's wait on customer feedback if that's going to be something they want from us. my assumption is if you are building with the sdk, you should have knowledge of the blockchain outside of Dfns. we are just demonstrating how to use Dfns keys to sign transactions. Everything else is not our responsibility to support.

Torakushi (dismissed):
LGTM,

I didn't test the synchronous signing though (but I trust you :))

ncornwell (dismissed):

ncornwell (dismissed):

ncornwell (approved):

PR #4, opened 2023-11-24
Nextjs example: load env from either `.env` or `.env.local`

Thibault2ss:
In the nextjs-delegated-example, have dotenv load environment variables from either `.env` or `.env.local`, so that either one would be properly parsed (included multiline variables).

### DIFF
--- a/examples/sdk/nextjs-delegated/.gitignore
+++ b/examples/sdk/nextjs-delegated/.gitignore
@@ -25,6 +25,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # local env files
+.env
 .env*.local
 
 # vercel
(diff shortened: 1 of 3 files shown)


drunkirishcoder (approved):
PR is 👍 

just a random thought, is it really necessary to support `.env.local`? I feel we should be more consistent on what we support across all examples. authv2 is already confusing enough, any variations we introduce will likely lead to more support issues.

ncornwell (approved):

Thibault2ss:
About your comment @drunkirishcoder , only reason we support `.env.local`, is that this is the standard in nextjs, cf [this](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#loading-environment-variables). So if the developer is a Nextjs fanboy, he'll probably use `.env.local` instinctively I suppose

PR #9, opened 2023-12-01
feat(ripple): dfns-1813: typescript-sdk integration for XRP

Torakushi:
add XRP support + example 

- [x] WAIT FOR XRP to be tier 2 first


packages/lib-ripple/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@dfns/lib-ripple",
```
drunkirishcoder: let's call this `lib-xrpl`, there was an old deprecated package called [ripple-lib](https://www.npmjs.com/package/ripple-lib), let's not cause any confusion what we integrate with.

packages/lib-ripple/README.md
```diff
@@ -0,0 +1,36 @@
+# Dfns Wallet for [Ripple](https://xrpl.org/references.html)
+
+Dfns wallet integration with Ripple xrpl.js (https://js.xrpl.org/l), makes working with xrp as simple and painless as possible.
```
drunkirishcoder:
```suggestion
Dfns wallet integration with xrpl.js (https://js.xrpl.org), makes working with Ripple as simple and painless as possible.
```

packages/lib-ripple/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      throw new DfnsError(-1, 'wallet not active', { walletId, status: res.status })
+    }
+
+    if (res.network !== 'Ripple' && res.network !== 'RippleTestnet') {
+      throw new DfnsError(-1, 'wallet is not bound to Ripple or RippleTestnet', { walletId, network: res.network })
+    }
+
+    // ripple adds the prefix "ed" to eddsa pubKey/private key to have
+    // the same bytes number as ecdsa pubKey.
+    const prefix = res.signingKey.scheme === 'EdDSA' ? 'ed' : ''
```
drunkirishcoder: can move this into the constructor.

packages/lib-ripple/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public get address(): string {
+    return deriveAddress(this.signingPubKey)
+  }
+
+  async signTransaction(transaction: Transaction): Promise<Transaction> {
+    const res = await this.dfnsClient.wallets.generateSignature({
+      walletId: this.metadata.id,
+      body: { kind: 'Transaction', transaction: encode(transaction) },
```
drunkirishcoder:
```suggestion
      body: { kind: 'Transaction', transaction: `0x${encode(transaction).toLowerCase()}` },
```

packages/lib-ripple/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    // ripple adds the prefix "ed" to eddsa pubKey/private key to have
+    // the same bytes number as ecdsa pubKey.
+    const prefix = res.signingKey.scheme === 'EdDSA' ? 'ed' : ''
+
+    return new DfnsWallet(res, options, prefix + res.signingKey.publicKey)
+  }
+
+  public get address(): string {
+    return deriveAddress(this.signingPubKey)
```
drunkirishcoder: should be able to just return `this.metadata.address`.

packages/lib-ripple/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      -1,
+      'cannot complete signing synchronously because this wallet action requires policy approval',
+      res
+    )
+  }
+}
+
+type WalletMetadata = GetWalletResponse
+
+export class DfnsWallet {
```
drunkirishcoder: ~looking into how to make the integration tighter, this is what we should do.~

~xrpl has a [wallet](https://github.com/XRPLF/xrpl.js/blob/main/packages/xrpl/src/Wallet/index.ts) abstraction. unfortunately it's not an interface. but we can still make this work by extending the class.~

```
export class DfnsWallet extends xrpl.Wallet
```

~we will override the following implementations, basically what you already have, with some minor changes to conform to their class.~

```
public get address(): string
public sign(transaction: Transaction, multisign?: boolean | string): {
    tx_blob: string
    hash: string
}
public getXAddress(tag: number | false = false, testnet = false): string
```

~should also override this, but throw not implemented error~

```
public verifyTransaction(): boolean
```

~nice thing with this design is then we can integrate with the xrpl much better like this~

```
const res = await client.submitAndWait({
    TransactionType: 'Payment',
    Account: wallet.address,
    Destination: "rBYtCQKxGTfFuob3hxSc8pEYddetT9CdDZ",
    Amount: '1',
  },
  {
    autofill: true,
    wallet,
  })
```

examples/libs/ripple/basic-tx/main.ts
```diff
@@ -0,0 +1,60 @@
...
+    // Replace with your desired receiver
+    Destination: "rBYtCQKxGTfFuob3hxSc8pEYddetT9CdDZ",
+    Amount: '100000',
+  }
+
+  // Choose the client you want.
+  const client = new Client('wss://testnet.xrpl-labs.com')
+  
+  // Connect the websocket client
+  await client.connect().then(async () => {
```
drunkirishcoder: don't use `then ... finally`. that's the old promise style before `async/await`.

packages/lib-ripple/index.ts
drunkirishcoder: nevermind, this would've worked except I just realized their `sign` function is not async. even though if they just made it async, it would work. `submitAndWait` is async.

that said, would still make sense I think to match the return type of `sign` to their function. minimize the differences between the examples they have and our example.

```
Promise<{
    tx_blob: string
    hash: string
}>
```

Torakushi: Yeah, I've seen their wallet abstraction, I did't think that it will be worth it. After seeing you're example, maybe it would have been more straight forward for the user. But unfortunately we can't use it :) 

I'll change the "sign" return. I'll keep in mind that we always want to keep things as close to the official library as possible.

examples/libs/ripple/basic-tx/main.ts
Torakushi: I miss my 'defer' statement in go :) 

It was only to call `disconnect` even in case of error

Will use a try catch instead

packages/lib-ripple/index.ts
Torakushi: right, forgot that now the address is in the metadata

README.md
```diff
@@ -177,8 +177,9 @@ Integrations with other blockchain platforms to make Dapp development frictionle
 
 * [BitcoinJS](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-bitcoinjs/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/bitcoinjs/)
 * [ethers.js v5](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs5/)/[v6](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-ethersjs6/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ethersjs)
-* [Tezos taquito](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-taquito/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/taquito)
+* [Ripple xrpl.js](https://github.com/dfns/dfns-sdk-ts/tree/m/packages/lib-xrpl/) and [examples](https://github.com/dfns/dfns-sdk-ts/tree/m/examples/libs/ripple)
```
drunkirishcoder: can we rename the example folder `xrpl` to match?

examples/libs/ripple/basic-tx/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Copy `.env.example` to a new file `.env` and set the following values,
+
+- `DFNS_API_URL` = `https://api.dfns.ninja`
+- `DFNS_APP_ID` = the `App ID` from above
+- `DFNS_APP_ORIGIN` = `http://localhost:3000`
+- `DFNS_CRED_ID` = the `Signing Key Cred ID` from above
+- `DFNS_PRIVATE_KEY` = the private key from the step 'generate a keypair', the newlines should not be a problem
+- `DFNS_AUTH_TOKEN` = the `authToken` from above, the value should start with `eyJ0...`
+- `DFNS_WALLET_ID` = a Dfns [wallet](https://docs.dfns.co/dfns-docs/api-docs/beta-wallets-api-and-nfts/create-wallet) ID
```
drunkirishcoder: `RIPPLE_WALLET_ID`

examples/libs/ripple/basic-tx/main.ts
drunkirishcoder: it's just a simple example, don't really care about error handling. all the examples we just assume they will work.

packages/lib-xrpl/index.ts
```diff
@@ -0,0 +1,77 @@
+import { DfnsApiClient, DfnsError } from '@dfns/sdk'
+import { Transaction, deriveAddress, encode } from 'xrpl'
```
drunkirishcoder: `deriveAddress` is no longer used.

packages/lib-xrpl/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  public get address(): string {
+    return this.metadata.address!
+  }
+
+  public async signTransaction(transaction: Transaction): Promise<{ tx_blob: string; hash: string }> {
+    const signatureResponse = await this.dfnsClient.wallets.generateSignature({
+      walletId: this.metadata.id,
+      body: { kind: 'Transaction', transaction: `0x${encode(transaction)}` },
```
drunkirishcoder:
```suggestion
      body: { kind: 'Transaction', transaction: `0x${encode(transaction).toLowerCase()}` },
```

packages/lib-xrpl/index.ts
```diff
@@ -0,0 +1,77 @@
+import { DfnsApiClient, DfnsError } from '@dfns/sdk'
+import { Transaction, encode } from 'xrpl'
+import { GetWalletResponse, GenerateSignatureResponse } from '@dfns/sdk/types/wallets'
+import { hashSignedTx } from 'xrpl/dist/npm/utils/hashes'
```
drunkirishcoder: let's try not to import from private package path (when possible), the way to get this properly seems to be add import `hashes` from `xrpl` above. then later you can call `hashes.hashSignedTx(signedTx)`

packages/lib-xrpl/index.ts
Torakushi: Why is that?

Indeed, `encode(transaction)` is upper-case, but that's no problem for the signer. Is it just to be consistent?

drunkirishcoder: yeah

drunkirishcoder (dismissed):

drunkirishcoder (approved):

PR #11, opened 2023-12-02
synchronous signing for evm libs

drunkirishcoder:
* remove polling from ethersjs and viem, signing is now completely synchronous.
* if the wallet is bound to an evm network, we will sign the original transaction instead of the hash of the transaction.
* marked the ethersjs constructor as `deprecated`, it still compiles. major problem is constructor can't be async, it makes doing some validation based on the wallet impossible inside the constructor. the static `init` is the preferred pattern and is what we use for all the other libs integrations.
* moved all the examples under a `libs` folder. the `sdk` examples are buried by the libs examples. now the sdk examples are a bit easier to see.
* made the `.env.example` files more clear.


packages/lib-ethersjs5/index.ts
```diff
@@ -1,93 +1,135 @@
...
 import { defineReadOnly, resolveProperties } from '@ethersproject/properties'
+import { toUtf8Bytes } from '@ethersproject/strings'
 import { computeAddress, serialize, UnsignedTransaction } from '@ethersproject/transactions'
 
-const sleep = (interval = 0) => new Promise((resolve) => setTimeout(resolve, interval))
-
 export type DfnsWalletOptions = {
   walletId: string
   dfnsClient: DfnsApiClient
+  /** @deprecated transaction signing is now synchronous. polling is deprecated. */
```
ncornwell, line 17: Can we remove these or are you trying to save backwards compatibility?

ncornwell (approved):

packages/lib-ethersjs5/index.ts
drunkirishcoder: trying to be a bit backwards compatible since this is probably the most used integrations lib. this `@deprecated` should show up in an IDE to remind people to migrate. in a future release we can remove them.

PR #20, opened 2024-01-04
feat: add aws kms key signer

ncornwell:
Add a key signer for AWS KMS

### DIFF
--- a/bin/publish.sh
+++ b/bin/publish.sh
@@ -16,6 +16,7 @@ declare -a packages=(
     "@dfns/sdk-keyexport-utils"
     "@dfns/sdk-keyimport-utils"
     "@dfns/sdk-keysigner"
+    "@dfns/sdk-awskmssigner"
     "@dfns/sdk-webauthn"
 )
 
--- a/bin/version_all.sh
+++ b/bin/version_all.sh
@@ -18,6 +18,7 @@ npm version --no-git-tag-version \
     --workspace @dfns/sdk-keyexport-utils \
     --workspace @dfns/sdk-keyimport-utils \
     --workspace @dfns/sdk-keysigner \
+    --workspace @dfns/sdk-awskmssigner \
     --workspace @dfns/sdk-webauthn \
     --include-workspace-root "$version"
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "packages/*"
       ],
       "dependencies": {
+        "@aws-sdk/client-kms": "^3.485.0",
         "@ethersproject/abstract-provider": "5.7.0",
         "@ethersproject/abstract-signer": "5.7.0",
         "@ethersproject/address": "5.7.0",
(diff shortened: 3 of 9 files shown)


packages/sdk-aws-kms-signer/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    const clientData = Buffer.from(
+      JSON.stringify({
+        type: 'key.get',
+        challenge,
+        origin: this.options.appOrigin,
+        crossOrigin: this.options.crossOrigin ?? false,
+      })
+    )
+
+    const client = new KMSClient({ region: this.options.awsRegion })
```
drunkirishcoder: I think this can be moved to the constructor, that will allow for connection reuse for multiple signing requests.

packages/sdk-aws-kms-signer/index.ts
```diff
@@ -0,0 +1,49 @@
+import { CredentialSigner, KeyAssertion } from '@dfns/sdk'
+import { toBase64Url } from '@dfns/sdk/utils'
+import { KMSClient, SignCommand, SigningAlgorithmSpec } from "@aws-sdk/client-kms";
+
+
+export class AwsKmsKeySigner implements CredentialSigner<KeyAssertion> {
+
+  constructor(
+    private options: {
+      kmsKeyId: string
```
Thibault2ss: Should this one be optional ? If it takes the default `AWS_REGION` in environment ? Or maybe the other way around: should we rather allow people to add the whole kms config ? having `kmsConfig?: KMSClientConfig` in the params instead of `awsRegion` (they might wanna pass aws creds in there as well 🤷 )

drunkirishcoder: same train of thought as above, maybe create a sub type that groups the kms key parameters for the sign command together.

```
kmsKeyConfig: {
  id: string
  algorithm: SigningAlgorithmSpec
}
```

packages/sdk-aws-kms-signer/index.ts
ncornwell: I updated this, let me know if it works given how the constructor operations flow or if I need to instantiate it directly in the constructor.

packages/sdk-aws-kms-signer/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@dfns/sdk-aws-kms-signer",
```
drunkirishcoder: nit, `sdk-awskmssigner` to have fewer dashes? otherwise looks good, test and :shipit:

packages/sdk-aws-kms-signer/index.ts
ncornwell: 👍  passing in the config is ideal, region is not required just copying from the docs 😅

packages/sdk-aws-kms-signer/package.json
ncornwell: Need to figure out our dash acceptance criteria 😂 , I've udpated

drunkirishcoder (approved):

packages/sdk-aws-kms-signer/package.json
drunkirishcoder: 😆

PR #24, opened 2024-01-08
fix: remove mediation required from webauthn sdk for better client support

ncornwell:


packages/sdk-webauthn/index.ts
```diff
@@ -17,14 +17,13 @@ export class WebAuthn implements CredentialSigner<Fido2Assertion>, CredentialSto
       rpId: string
       timeout?: number
     }
-  ) {}
+  ) { }
```
Thibault2ss, line 20: is that your linter ?

Thibault2ss (approved):

packages/sdk-webauthn/index.ts
ncornwell: Probably, we should add our standard lint/pre-commit check config

drunkirishcoder (approved):

PR #39, opened 2024-01-29
feat: add web based key signer

ncornwell:


C0x41lch0x41 (dismissed):

packages/sdk-websigner/index.ts
```diff
@@ -0,0 +1,77 @@
+import * as crypto from 'crypto'
+import { CredentialSigner, KeyAssertion } from '@dfns/sdk'
+import { toBase64Url } from '@dfns/sdk/utils'
+
+export class WebKeySigner implements CredentialSigner<KeyAssertion> {
```
Thibault2ss: nit: is the name "Web" clear that it's meant for "client-side" (browser-side)? Maybe `BrowserKeySigner` + package `sdk-browser-keysigner` 🤷

ncornwell: Updated

packages/sdk-browser-keysigner/index.ts
```diff
@@ -0,0 +1,77 @@
...
+    const clientData = JSON.stringify({
+      type: 'key.get',
+      challenge,
+      origin: this.options.appOrigin,
+      crossOrigin: this.options.crossOrigin ?? false,
+    })
+
+
+    const rawSignature = await crypto.subtle.sign(
+      { name: 'ECDSA', hash: { name: 'SHA-256' } },
```
djin-dfns: is `ECDSA` the only thing can be supported?

examples/sdk/browser-keysigner/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "sdk-browser-keysigner-example",
```
djin-dfns, line 2: does it need to be a separate package or can this just go into the keysigner package? given it has 0 external lib dependency?

packages/sdk-browser-keysigner/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@dfns/sdk-browser-keysigner",
```
djin-dfns: does it need to be a separate package or can this just go into the keysigner package? given it has 0 external lib dependency?

examples/sdk/browser-keysigner/index.ts
```diff
@@ -0,0 +1,60 @@
+import { DfnsApiClient } from '@dfns/sdk'
```
djin-dfns, line 2: nit: given this is meant to be used in the browser, would a browser based example showcase it better?

ncornwell: Yes!

packages/sdk-browser-keysigner/package.json
ncornwell: done

packages/sdk-browser-keysigner/index.ts
ncornwell: Yes more protocols can be support, more configuration options will be needed though

packages/sdk-browser-keysigner/package.json
Thibault2ss: mh though the keysigner package is meant for backend side, no ? It uses nodejs `crypto` module, not the navigator crypto one

ncornwell: If you don't use the AsymetricSigner then it doesn't matter

packages/sdk-browser-keysigner/index.ts
djin-dfns: we already have the `algorithm` parameter that's not being used here. maybe we can change it's type from `string` to

```
export type CryptoSubtleAlgorithm = Parameters<typeof crypto.subtle.sign>[0]
```

```
  constructor(
    private options: {
      privateKey: CryptoKey
      credId: string
      appOrigin: string
      crossOrigin?: boolean
      algorithm: CryptoSubtleAlgorithm
    }
  ) { }
```

packages/sdk-browser-keysigner/package.json
djin-dfns: yeah as long as it's not bringing in unnecessary dependencies I think it's ok, since both are just key signers. ideally if we can have a single key signer that works in both environments, that would be even better, but probably not worth the effort.

ncornwell: In theory node crypto is faster than browser subtle crypto, keep 2 for now 😅

packages/sdk-browser-keysigner/index.ts
ncornwell: EdDSA isn't broadly supported yet, I'm going to remove `algorithm` from the params as we can figure it out from the key (`this.options.privateKey.algorithm.name`) and we have limited hash/curve support.

djin-dfns: what about RSA? did we say we want to encourage RSA because people get super confused about ECDSA for action signing and ECDSA in mpc?

djin-dfns (approved):

packages/sdk-browser-keysigner/index.ts
ncornwell: I'll think on that one, RSA can also be a bit confusing, ECDSA I guess does have different curves + hashing algorithms though we would only expect 1.

packages/sdk-browser-keysigner/package.json
Thibault2ss: Ok I guess what's weird is that someone can install that module and use one thing, but not another or it'll fail... a bit weir I think

PR #46, opened 2024-02-09
feat: viem2 example

ncornwell:


djin-dfns:
looks like [alchemy aa sdk](https://github.com/alchemyplatform/aa-sdk/tags) is going to upgrade to viem2 soon, maybe we can just wait for that and upgrade our viem integration.

PR #53, opened 2024-02-27
Add polkadot integration

adriendfns:
Add polkadot integration


packages/lib-polkadot/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    const response = await this.dfnsClient.wallets.generateSignature({
+      walletId: this.walletId,
+      body: { kind: 'Transaction', transaction: data, },
+    })
+
+    assertSignResponseSuccessful(response)
+
+    return u8aToHex(
```
Torakushi: Is there a reason to use this instead of Buffer.Concat?

From what I've seen, u8aFunction are useful when you deal with exotic `U8aLike`. Here you only have Buffer

(Maybe I am missing something)

packages/lib-polkadot/polkadot-signer.ts
```diff
@@ -0,0 +1,15 @@
+import { Signer, SignerResult } from '@polkadot/api/types'
+import { SignerPayloadRaw } from '@polkadot/types/types/extrinsic'
+import { DfnsWallet } from './index'
+
+export class PolkadotSigner implements Signer {
+    // Id we increment for each signature
+    private id: number
```
Torakushi: Just curious, what is it for?

packages/lib-polkadot/index.ts
adriendfns: you are totally correct, It is because I have reused the existing code: https://github.com/dfnsco/blockchain-integrations/blob/m/polkadot-wallet/src/polkadot-signer.ts#L28

And I did not have a closer look, I now have simplified this part a lot

packages/lib-polkadot/polkadot-signer.ts
adriendfns: The type `SignerResult` asks us to return a number. It feels safe to return a different one for each request. 

We can see an example here: https://github.com/polkadot-js/apps/blob/master/packages/react-signer/src/signers/AccountSigner.ts#L29C29-L29C30

I am afraid I don't know more about the underlying logic

ncornwell (dismissed):

packages/lib-polkadot/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  public async broadcast(transaction: string) {
+    const response = await this.dfnsClient.wallets.broadcastTransaction({
+      walletId: this.walletId,
+      body: { kind: 'Transaction', transaction, },
+    })
+
+    return response
+  }
```
djin-dfns: let's drop broadcast. the sdk integration is meant to be working at the signature level, including chains we don't have providers for. broadcast is not part of the sdk integration.

packages/lib-polkadot/polkadot-signer.ts
```diff
@@ -0,0 +1,17 @@
+import { Signer, SignerResult } from '@polkadot/api/types'
+import { SignerPayloadJSON } from '@polkadot/types/types/extrinsic'
+import { DfnsWallet } from './index'
+
+export class PolkadotSigner implements Signer {
```
djin-dfns: `DfnsWallet` can implement `Signer` interface, we don't need this extra class.

packages/lib-polkadot/index.ts
adriendfns: removed

packages/lib-polkadot/polkadot-signer.ts
adriendfns: done

packages/lib-polkadot/index.ts
```diff
@@ -0,0 +1,81 @@
...
+    return new DfnsWallet(res.address!, options)
+  }
+
+  // Implementation of Signer
+  async signRaw(raw: SignerPayloadRaw): Promise<SignerResult> {
+    const signature = await this.generateSignature(raw.data, raw.address)
+    return { id: ++this.id, signature: signature }
+  }
+
+  public async generateSignature(data: string, address: string) {
```
djin-dfns: make this private? would prefer `signRaw` instead of trying to call this directly. gives us more flexibility in future refactoring.

examples/libs/polkadot/broadcast/main.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const signerPayload = registry.createType('SignerPayload', unsigned, { version: unsigned.version })
+
+  // Broadcast transaction
+  const response = await dfnsClient.wallets.broadcastTransaction({
+    walletId: polkadotWalletId,
+    body: { kind: 'Transaction', transaction: signerPayload.toHex() },
+  })
+  console.log(`Transaction submitted with hash ${response.txHash}`)
+}
```
djin-dfns: let's drop the broadcast example too. instead, we will strip this down as much as possible and add to our docs (like [here](https://docs.dfns.co/d/api-docs/wallets/broadcast-transaction-from-wallet/solana-broadcast-transaction)).

examples/libs/polkadot/basic-tx/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "polkadot-basic-tx",
+  "version": "1.0.0",
+  "main": "main.ts",
+  "dependencies": {
+    "@polkadot/api": "10.11.3",
+    "@dfns/lib-polkadot": "file:../../../packages/lib-polkadot"
+  },
```
djin-dfns, line 9: missing `dotenv` as a dependency.

djin-dfns (approved):

PR #58, opened 2024-03-13
Add create method for BrowserKeySigner

C0x41lch0x41:
It is now possible to use DfnsAuthenticator.register method with
BrowserKeySigner (direct auth) or call BrowserKeySigner.create directly
(delegated auth)

```javascript
const browserSigner = new BrowserKeySigner({
  credId: credId,
  keyPair: keyPair,
  appOrigin: appOrigin
})
const dfnsAuth = new DfnsAuthenticator({
  appId: appId,
  baseUrl: baseUrl,
  signer: browserSigner,
})
await dfnsAuth.register({ username, orgId, registrationCode })
```


packages/sdk-keysigner/index.ts
```diff
@@ -78,6 +78,62 @@ export class BrowserKeySigner implements CredentialSigner<KeyAssertion> {
...
 
+  toHex = (buffer: ArrayBuffer): string => {
+    const view = new Uint8Array(buffer)
+    let hexString = ""
+    for (const byte of view) {
+      const hexByte = byte.toString(16)
+      hexString += hexByte.padStart(2, "0")
+    }
+    return hexString.toLowerCase()
+  }
```
Thibault2ss: nit: maybe this can be extracted to a util outside of the class ?

packages/sdk-keysigner/index.ts
```diff
@@ -78,6 +78,62 @@ export class BrowserKeySigner implements CredentialSigner<KeyAssertion> {
...
+    if (algorithm == 'ECDSA') {
+      rawSignature = await crypto.subtle.sign(
+        { name: 'ECDSA', hash: { name: 'SHA-256' } },
+        this.options.keyPair.privateKey,
+        new TextEncoder().encode(credInfoFingerprint)
+      )
+    } else {
+      throw new Error(`${algorithm} is not supported`)
+    }
+    const signature = this.rawSignatureToAns1(new Uint8Array(rawSignature))
```
Thibault2ss: fyi (not directly linked to this PR) I'm not sure how reliable/accurate the `rawSignatureToAns1` function actually is. I have never digged really. Noah took it from Cody's code I think

ncornwell: We could switch to `KJUR/jsrsasign` though I think keeping dependencies light can always be helpful.

packages/sdk/utils/base64.ts
```diff
@@ -17,3 +17,8 @@ export const fromBase64Url = (encoded: string): Buffer => {
   encoded = encoded.replace(/\-/g, '+').replace(/_/g, '/')
   return Buffer.from(encoded, 'base64')
 }
+
+export const toBase64 = (buffer: ArrayBuffer): string => {
+  const view = new Uint8Array(buffer);
+  return btoa(String.fromCharCode(...view));
```
djin-dfns: can we not use `Buffer.toString` here? or should we switch to `btoa` everywhere? prefer to be consistent on b64 handling.

packages/sdk-keysigner/index.ts
Thibault2ss: Yeah we can keep this, it's just that by reading the function, without digging on the actual spec, I cannot tell if the function is actually good or not. I guess empiric proof is good as well, we'll see if some customer complains sometime :)

djin-dfns:
We have a bunch of conversation ongoing to add more auth functionality targeting the browser. maybe it would make sense to pull all the browser specific things into a new package `sdk-browser`, we can move this one the browser signer, webauthn signer and the upcoming iframe signer into that package. seems to be a slightly better way to organize?

packages/sdk-keysigner/index.ts
djin-dfns: also we already take a dependency on browser compatible [buffer](https://www.npmjs.com/package/buffer) which I can do this. unless we don't want to have that dependency anymore.

packages/sdk-keysigner/index.ts
```diff
@@ -78,6 +78,62 @@ export class BrowserKeySigner implements CredentialSigner<KeyAssertion> {
...
+    for (const byte of view) {
+      const hexByte = byte.toString(16)
+      hexString += hexByte.padStart(2, "0")
+    }
+    return hexString.toLowerCase()
+  }
+
+  exportPublicKeyInPemFormat = async (key: CryptoKeyPair): Promise<string> => {
+    const exported = await crypto.subtle.exportKey('spki', key.publicKey)
+    const pem = `-----BEGIN PUBLIC KEY-----\n${toBase64(exported)}\n-----END PUBLIC KEY-----`
```
djin-dfns: I believe, we also need to break the base64 encoding with new line every 64 characters to follow strict PEM formatting.

packages/sdk-keysigner/index.ts
```diff
@@ -78,6 +78,62 @@ export class BrowserKeySigner implements CredentialSigner<KeyAssertion> {
...
+    const signature = this.rawSignatureToAns1(new Uint8Array(rawSignature))
+    const attestationData = JSON.stringify({
+      publicKey: publicKeyPem,
+      signature: this.toHex(signature) 
+    })
+
+    return {
+      credentialKind: 'Key',
+      credentialInfo: {
+        credId: this.options.credId,
```
djin-dfns: this feels a bit weird, old service account key signer only signs, and it needs the credId of the already created credential so the signing can be verified on the server side.

but here, this credential hasn't been created yet. they don't have a credId. are we allowing them to pick their own credId?

packages/sdk-keysigner/index.ts
C0x41lch0x41: Yes this is pretty straightforward and will not require updates so I'd advocate to keep it in the code without adding an unnecessary dependency.
I tested it the complete flow and was able to go from Registration Code to Registration with Key Credential and Login so that seems to work as expected.

packages/sdk-keysigner/index.ts
C0x41lch0x41: Yeah I don't have a strong opinion about this. In my testing example I generate a random credId and pass it to the constructor. I thought we could let the customer choose a credId. But since it is possible to pass a name to the credential that does not make much sense. I'll create a random one.

packages/sdk/utils/base64.ts
C0x41lch0x41: Yes I tried to avoid the Buffer dependency since it does not exists natively on the browser side, what do you think?
I see it is in the `sign` method though but I can change it.

packages/sdk-keysigner/index.ts
C0x41lch0x41: I wanted to avoid this dependency (less dependency = less supply chain attack surface) but if we think it would be too difficult to maintain without it I can add it.

packages/sdk-keysigner/index.ts
C0x41lch0x41: I put the function in utils and split to 64 bytes per line. Feel free to resolve the conversation if that's ok for you.

packages/sdk-keysigner/index.ts
C0x41lch0x41: I put the function in utils.

packages/sdk/utils/base64.ts
C0x41lch0x41: I am not a fan of having the buffer dependency for the browser but since we support both browser and node code keeping it seems cleaner for now. I removed `btoa` usage and cleaned base64 code.

packages/sdk-keysigner/index.ts
C0x41lch0x41: Closing this. Feel free to reopen it if you have an additional comment.

packages/sdk-keysigner/index.ts
C0x41lch0x41: If credId is not passed to the constructor it is randomly generated.

packages/sdk-keysigner/index.ts
C0x41lch0x41: I moved this to a utils function

packages/sdk/utils/base64.ts
djin-dfns: yeah I'm not against switching over to `btoa` across the board. it's your call if you think it's not worth taking a dependency.

packages/sdk/utils/bigint.ts
```diff
@@ -0,0 +1,13 @@
...
+        if (minValue[i] === 0) {
+            continue
+        }
+        if (minValue[i] > 0x7f) {
+            return new Uint8Array(minValue.slice(i - 1))
+        }
+        return new Uint8Array(minValue.slice(i))
+    }
+    return new Uint8Array([0])
+}
```
djin-dfns: nit: can you have your editor to add new line at end of file. 😄 personal pet peeve of mine.

packages/sdk-keysigner/index.ts
djin-dfns: thinking more about this, maybe we do want them to pick the `credId`, or have a way to generate a "stable" `credId`. one problem I'm thinking generating a random one is, if they call the `create` multiple times, they can potentially create the credential over and over.

and they have no way of knowing the right `credId` to configure the sign with in their code if they are not the ones choosing. that's another usability problem I think.

anyway, sorry I think what you had before is correct. they must provide a `credId`. also minor detail, `credId` is base64url encoded, we should ensure it's base64url encoded here too to be consistent I think (even though it doesn't change anything functionally)

C0x41lch0x41: > anyway, sorry I think what you had before is correct. they must provide a credId

Yes so with my approach either they don't provide a credId and it randomly generated or they provide one and we use it. If it is randomly generated they can get it from the returned object `credentialInfo.credId`. 
Are you with this approach?

> credId is base64url encoded, we should ensure it's base64url encoded here too to be consistent I think

I looked into that and it is actually not trivial to make sure a string is base64 or not. All the ascii string once padded can actually be base64 decoded. I am afraid the check can actually introduce some uncontrolled edge cases.

packages/sdk/utils/bigint.ts
C0x41lch0x41: Lol fair enough. I pushed a new version with new lines :-).

packages/sdk-keysigner/index.ts
djin-dfns: yeah maybe we base64 it when we generate one, otherwise just leave it as is. 👍

djin-dfns (approved):

PR #62, opened 2024-03-19
action signing with app relying party

djin-dfns:


packages/sdk/signer.ts
```diff
@@ -56,5 +79,5 @@ export type SecondFactorAssertion = KeyAssertion | Fido2Assertion | TotpAssertio
 export type CredentialAssertion = KeyAssertion | Fido2Assertion | PasswordAssertion | TotpAssertion
 
 export interface CredentialSigner<T extends CredentialAssertion = FirstFactorAssertion> {
-  sign(challenge: string, allowCredentials: { key: AllowCredential[]; webauthn: AllowCredential[] }): Promise<T>
+  sign(challenge: UserActionChallenge): Promise<T>
```
djin-dfns, line 85: @Thibault2ss @C0x41lch0x41 I want to make this change to the signer interface. two things we are adding here

* `rpid`, so we no longer need to configure webauthn/passkeys with rpid on the client side
* `challengeIdentifier`, this is needed for the iframe signer to display inside iframe exact what is being signed.

what do you guys think? it will be breaking for people who's using the delegated client/server pattern because they are calling `webauthn.sign` directly, otherwise, it's in theory hidden from the user if they are using our sdk wrappers that completely hides this.

C0x41lch0x41: Yes I like the idea. Specifying the RP ID on the client side always felt weird to me.
Could you push the code that integrates this change in the signers (webauthn, asymmetric and browser) as well as the examples so that I am sure I have a full understanding of the scope of this?

Also I am not sure I understand how this interface is declared and why `sign` is there and not `create`? 

Last thing, do you think we can have two versions of the `sign` function one with `challenge:string, allowCredentials` args and one with `UserActionChallenge` with the later being a wrapper of the former so that it's not a breaking change?

djin-dfns: pushed up the changes for the signers. will update all the examples. it's possible to do an overload but because we are moving the WebAuthn signer anyway, it's a "breaking" change already, might as well drop all traces of `rpid` and `origin`. that's my preference.

packages/sdk-react-native/index.ts
```diff
@@ -97,23 +93,20 @@ class AndroidPasskeys implements InnerSigner {
...
-  constructor(private options: PasskeysOptions) {}
-
-  async sign(
-    challenge: string,
-    allowCredentials: { key: AllowCredential[]; webauthn: AllowCredential[] }
-  ): Promise<Fido2Assertion> {
-    const request = {
-      challenge: b64UrlSafeToStandard(challenge),
-      allowCredentials: allowCredentials.webauthn.map(({ id, type, transports }) => ({
+class iOS implements CredentialSigner<Fido2Assertion>, CredentialStore<Fido2Attestation> {
```
Thibault2ss: "IOS" and "Android" class names feel too generic 🤔 If I import that from a lib, I'm like "Am i importing whole IOS ? 😅 "

examples/sdk/auth-delegated/mobile/package.json
```diff
@@ -10,7 +10,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "@dfns/sdk-react-native": "0.3.3",
+    "@dfns/sdk-react-native": "0.4.0-alpha.2",
```
C0x41lch0x41: I this `alpha.2` expected for a dfns package?

packages/sdk-react-native/index.ts
djin-dfns: they are private classes, not exported. user only will be using the `PasskeysSigner`. but I can rename them back.

C0x41lch0x41 (dismissed):

examples/sdk/auth-delegated/mobile/package.json
djin-dfns: nope. let me fix that

packages/sdk-react-native/index.ts
Thibault2ss: Oh fine then I didn't realize they were private classes

Thibault2ss:
image

ncornwell (approved):

C0x41lch0x41 (approved):

PR #65, opened 2024-03-22
browser keysigner with examples

C0x41lch0x41:
- Standalone browsersigner SDK and examples:
- Auth delegated web example for browser keysigner

Added the ability to choose between webauthn and key credential
Key credential is not advised to use at is as the private key is in the 
browser so it is not a secure solution


examples/sdk/auth-delegated/web/src/pages/Register.tsx
```diff
@@ -1,20 +1,30 @@
 import { WebAuthn } from '@dfns/sdk-webauthn'
+import { BrowserKeySigner } from '@dfns/sdk-browsersigner'
```
Thibault2ss: I think there was discussions around gathering any features meant for browsers in the same `@dfns/sdk-browser` package or smth ? Where we would have this `BrowserKeySigner` + `WebauthNSigner`, since both are meant to be used in browsers

examples/sdk/auth-delegated/web/src/pages/Register.tsx
```diff
@@ -1,20 +1,30 @@
...
     try {
       setLoading(true)
       event.preventDefault()
 
       const formData = new FormData(event.currentTarget)
+      const credtype = formData.get('credtype') as string
+      let isWebAuthn = true
+      if (credtype === 'key') {
+        isWebAuthn = false
+      }
```
Thibault2ss: nit
```suggestion
      const isWebAuthn = formData.get('credtype') !== 'key'
```

examples/sdk/auth-delegated/web/src/pages/Register.tsx
```diff
@@ -33,9 +43,32 @@ export default function Register(): JSX.Element {
...
+        // In a production environment they key should be protected 
+        // and loaded securely in the browser
+        const generatedKeyPair = await crypto.subtle.generateKey(
+          { name: 'ECDSA', namedCurve: 'P-256' },
+          true,
+          ['sign', 'verify']
+        )
+        // Here the private key is set as a session variable
+        // key will not exists upon logout
+        setKeyPair(generatedKeyPair)
```
Thibault2ss: I think the key will not even exist upon refresh of the page ? I don't think our app context is persisted in localstorage in this example

packages/sdk-browsersigner/index.ts
```diff
@@ -0,0 +1,104 @@
...
+import { toBase64Url, toHex, exportPublicKeyInPemFormatBrowser, rawSignatureToAns1, generateRandom } from '@dfns/sdk/utils'
+import { Buffer } from 'buffer'
+
+export class BrowserKeySigner implements CredentialSigner<KeyAssertion> {
+  constructor(
+    private options: {
+      keyPair: CryptoKeyPair
+      credId?: string
+      appOrigin: string
+      crossOrigin?: boolean
```
Thibault2ss: maybe we can ditch those fields ? Since
- `origin` should not be needed anymore for a Key Cred: we have a pending [PR](https://github.com/dfnsco/dfns/pull/1269) for that
- `crossOrigin`: not needed as well, doesn't make sense for key creds. It was only added that field to be "ISO webauthN", I believe that field is added in webauthN `clientData` in assertions. Afaiu, webauthN sets it to `true` if the cred was used in a document that's a child of another site (iframe/popup), which has another origin ([in the spec](https://www.w3.org/TR/webauthn-3/#dictionary-client-data) there's also another field `topOrigin` that can be added to clientData). But I don't think we actually verify this [during assertion validation](https://github.com/dfnsco/dfns/blob/8dd658c1ab3d19d217db92928b101dc31f2b06f2/services/auth/dockers/security-auth-management/src/utils/webauthn.ts#L179). Anyways for Key creds I think it doesn't make sense since those key creds can be used anywhere

examples/sdk/auth-delegated/web/src/pages/Register.tsx
djin-dfns: yeah would like to consolidate all the browser specific functionalities into one package and avoid more small packages. 👍

djin-dfns:
In terms of the example, I wonder if it's better to have a separate standalone example (a copy of the nextjs one maybe).

* We want to discourage the use of this signer. but showcase this next to WebAuthn I feel in a way does the opposite of that.
* Also we want to keep the example code as simple as possible to understand given how much of a support burden it's been so far. Now adding the branching to both register and sign introduces more "complexity" when someone new looks at the code, figuring out which path to follow.

examples/sdk/auth-delegated/web/src/pages/Register.tsx
C0x41lch0x41: Yes makes sense. I'll do it.

examples/sdk/auth-delegated/web/src/pages/Register.tsx
C0x41lch0x41: Yes that's right. It is not the goal of this example (which is about using the sdk) but it highlights that securing the key is not trivial at all.

C0x41lch0x41:
> In terms of the example, I wonder if it's better to have a separate standalone example (a copy of the nextjs one maybe).
> 
> * We want to discourage the use of this signer. but showcase this next to WebAuthn I feel in a way does the opposite of that.
> * Also we want to keep the example code as simple as possible to understand given how much of a support burden it's been so far. Now adding the branching to both register and sign introduces more "complexity" when someone new looks at the code, figuring out which path to follow.

Yes good call. It did not feel right to implement browserkeysigner which is not secure alongside a very secure one like webauthn. I wanted to go for the shortest path though so still went for it. 
But with this extra feedback I feel more confident we should not mix the two in the examples. I'll create a separate example then. Thanks!

packages/sdk-browsersigner/index.ts
C0x41lch0x41: I removed `crossOrigin`. I'll create a separate PR to remove origin once we deploy the corresponding backend PR (it's easier for testing to keep it for now).

package.json
```diff
@@ -20,8 +20,8 @@
     "@ethersproject/properties": "5.7.0",
     "@ethersproject/strings": "5.7.0",
     "@ethersproject/transactions": "5.7.0",
+    "@meshsdk/core": "^1.5.11-beta.6",
```
djin-dfns: let's not pick up beta releases.

packages/sdk-browsersigner/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "@dfns/sdk-browsersigner",
```
djin-dfns: nit: let's just go with `sdk-browser`, we can include more helpers besides signers in the future maybe.

djin-dfns (dismissed):

package.json
C0x41lch0x41: Yes not ideal. That's what `npm audit fix` came up with :-P. Let me see if I can do better.

djin-dfns (approved):

PR #74, opened 2024-04-08
fix: switch auth client to newly generated client

Thibault2ss:
- switch `AuthClient` to the newly generated client (from `/generated/auth/client`)
- update examples to use the newly named auth methods, instead of the deprecated names
- create a `EmptySigner` class
- some other api changes reflected after SDK generation:
  - networks client: `maxFee` -> `maxFeePerGas`
  - wallet client:  `maxPriorityFee` -> `maxPriorityFeePerGas` change
  - permissions client: `dateCreated` and `dateUpdated` are not optional + `Auth:Creds:Code:Create` permission added
  - wallet client: added optional `priority` field

### DIFF
--- a/packages/sdk/signer.ts
+++ b/packages/sdk/signer.ts
@@ -86,9 +86,3 @@ export type CredentialAssertion = KeyAssertion | Fido2Assertion | PasswordAssert
 export interface CredentialSigner<T extends CredentialAssertion = FirstFactorAssertion> {
   sign(challenge: UserActionChallenge): Promise<T>
 }
-
-export class EmptySigner implements CredentialSigner {
-  async sign(): Promise<never> {
-    throw new DfnsError(-1, 'A signature is required, but an EmptySigner cannot fulfill it.')
-  }
-}
--- a/packages/sdk/signer.ts
+++ b/packages/sdk/signer.ts
@@ -1,3 +1,5 @@
+import { DfnsError } from './dfnsError'
+
 export type CredentialFactor = 'first' | 'second' | 'either'
 
 export type CredentialKind = 'Key' | 'Fido2' | 'Password' | 'Totp' | 'RecoveryKey'
@@ -83,4 +85,10 @@ export type CredentialAssertion = KeyAssertion | Fido2Assertion | PasswordAssert
 
 export interface CredentialSigner<T extends CredentialAssertion = FirstFactorAssertion> {
   sign(challenge: UserActionChallenge): Promise<T>
+}
+
+export class EmptySigner implements CredentialSigner {
+  async sign(): Promise<never> {
+    throw new DfnsError(-1, 'A signature is required, but an EmptySigner cannot fulfill it.')
+  }
 }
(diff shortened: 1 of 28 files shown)


Torakushi (dismissed):
LGTM

djin-dfns:
can you include the old generator as well, it will mark all the old types as deprecated and we can remove them later.

packages/sdk/signer.ts
```diff
@@ -84,3 +86,9 @@ export type CredentialAssertion = KeyAssertion | Fido2Assertion | PasswordAssert
 export interface CredentialSigner<T extends CredentialAssertion = FirstFactorAssertion> {
   sign(challenge: UserActionChallenge): Promise<T>
 }
+
+export class EmptySigner implements CredentialSigner {
+  async sign(): Promise<never> {
+    throw new DfnsError(-1, 'A signature is required, but an EmptySigner cannot fulfill it.')
+  }
+}
```
djin-dfns: why do we need this?

Thibault2ss: It's a helper placeholder, to be able to use the auth client for the un-authed endpoints, eg:
```ts
const dfnsClient = new DfnsApiClient({
    appId: process.env.DFNS_APP_ID!,
    baseUrl: process.env.DFNS_API_URL!,
    authToken: '',
    signer: new EmptySigner(),
  })
```

I was thinking eventually we don't need an additional class ("`BaseAuthApi`") to re-expose un-authed endpoints, maybe that simplifies the number of things people have to know to interact with our API (and that we have to do)

Thibault2ss: other thought -- unrelated -- maybe we should make `baseUrl` optional (by default it's gonna be the prod url)

djin-dfns: in that case wouldn't it be better if `authToken` and `signer` are nullable?

re: `baseUrl` I kinda like being explicit, feels less likely that they will be hitting the wrong api (ninja vs prod).

Thibault2ss: yeah, we can also make auth token and signer nullable, but I think original reason we did not do that, is because since 98% of cases you need those to use the api, we did not wanna wrongfully push people to "not fill those", since they wouldn't see the TS warnings. Though that's just for TS, not JS. So I'd be ok to make them nullable (not optional), and raise an error if authToken/signer is needed but not specified

re: `baseUrl` yeah ok. I think I'm thinking more towards PLG when people are just onboarded to prod. Would be nicer for them if they don't need to think about the api endpoint, just download the SDK, add an `appId` and that's it

djin-dfns: ok, let's make them nullable 👍

djin-dfns: if long term we still want to retire ninja, then I guess defaulting to prod would make sense eventually. that will be the only possible env externally at some point.

Thibault2ss: ok updated options as optional and so removed the unecessary `EmptySigner`

djin-dfns (approved):

PR #77, opened 2024-04-16
add recovery credential

djin-dfns:


packages/sdk-browser/keys/password.ts
```diff
@@ -0,0 +1,105 @@
...
+        name: 'PBKDF2',
+        salt,
+        iterations: 100000,
+        hash: 'SHA-256',
+      },
+      passwordKey,
+      { name: 'AES-GCM', length: 256 },
+      true,
+      ['wrapKey', 'unwrapKey']
+    )
```
Thibault2ss, line 82: nit: this part is used in both `.create()` and `.sign()`, it could be extracted in a method eg `.deriveWrappingKeyFromPassword(password)`

packages/sdk-browser/keys/password.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    const privateKey = await crypto.subtle.unwrapKey(
+      'pkcs8',
+      wrappedKey,
+      wrappingKey,
+      {
+        name: 'AES-GCM',
+        iv: iv,
+      },
+      { name: 'ECDSA', namedCurve: 'P-256' },
```
Thibault2ss, line 92: nit: maybe some const are worth extracting in their own const, and then references in the other methods ? eg
```ts
private keyPairConf = { name: 'ECDSA', namedCurve: 'P-256' }
private wrappingKeyConf = { name: 'AES-GCM', length: 256 }
```

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+        break
+      case 'sign':
+        break
+    }
+  } catch (error: any) {
+    e.ports[0].postMessage({ error })
+  }
+}
+
+export class RecoveryKeySigner {
```
Thibault2ss, line 101: Should we call it as such, since it's a "browser-side generated recovery key" ? (customers could also create a Recovery Key from another type of client eg IOS / Android). Although maybe the fact that this is inside the `@dfns/sdk-browser` is enough 🤔 

Also, should this class implement `CredentialSigner` ?

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(private readonly worker: Worker) {}
+
+  public create(recoveryKey: string, challenge: RecoveryCredentialChallenge): Promise<RecoveryKeyAttestation> {
+    return postAsync(this.worker, { kind: 'create', recoveryKey, challenge })
+  }
```
Thibault2ss, line 106: I'm less and less a fan of the choice of name `.create()` for the signer interface 😅 it's a bit confusing. I think it should be something like `.signRegistrationChallenge`. 

Reason why we added `.create()` originally was that we created the `WebauthN` signer first, and with webauthn the creation+signature of a challenge is done in one go. But since we then separated a `CredentialStore` class, which clearly separates the storing ability.

And this RecoveryKey class is not really a `CredentialStore`, it's just a `CredentialSigner`.

I think one way we could resolve that is 
- change the `CredentialSigner` interface to expose both `sign` (aka "`signUserActionChallenge`") and `signRegistrationChallenge`
- remove the common `CredentialStore` interface, since it's only used by WebauthNSigner currently, and if one day we actually define a different cred store, its `.create` method would not take in the same arguments as the `WebauthnSigner.create(challenge: UserRegistrationChallenge)`, since it probably wouldn't need the challenge in order to create a new key (this is particular to webauthn)
- `WebauthNSigner` would implement just `CredentialSigner`, and its `.signRegistrationChallenge` should just be a relay to call the custom `.create` method (that we would leave as a separate method, to make clearer that the webauthN signer actually creates and stores a new key)

examples/sdk/auth-delegated/web/src/background/recover.ts
Thibault2ss, line 1: nit: should we call this `src/workers/recover.ts` ? to make it more obvious that this is meant to be used in a a worker  ?

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch (error: any) {
+    e.ports[0].postMessage({ error })
+  }
+}
+
+export class RecoveryKeySigner {
+  constructor(private readonly worker: Worker) {}
+
+  public create(recoveryKey: string, challenge: RecoveryCredentialChallenge): Promise<RecoveryKeyAttestation> {
+    return postAsync(this.worker, { kind: 'create', recoveryKey, challenge })
```
Thibault2ss, line 105: nit: should we call `kind` -> `messageKind` so it's more explicit what this represents ?

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+    encryptedPrivateKey: key.encryptedPrivateKey,
+  }
+}
+
+export const handleRecoveryKeySignerMessage = async (e: MessageEvent<RecoveryKeySignerMessage>) => {
+  try {
+    switch (e.data.kind) {
+      case 'create':
+        const attestation = await create(e.data.recoveryKey, e.data.challenge)
+        e.ports[0].postMessage({ result: attestation })
```
Thibault2ss, line 91: One thing that I'm not sure about (not clear to me at least), is that `e.ports[0]` really is the port of the new `MessageChannel()` you instantiated on the other side, rather than the default port associated to the worker 🤔 

[This stackoverflow](https://stackoverflow.com/questions/30741686/spawning-a-shared-worker-in-a-dedicated-worker/30796101#30796101) (though a bit old) seems to suggest that the port needs to be passed in the event data itself to be used on the other side. However the [article](https://advancedweb.hu/how-to-use-async-await-with-postmessage/) you found seems to suggest so:
> To attach a port to the postMessage call, use the second argument: postMessage(data, [channel.port2]). To use this port on the other end, use event.ports[0].postMessage().

The [definition](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage) of the `worker.postMessage()` just says that this second argument is to pass objects which ownership will change from main thread to worker thread (unrefed on the main thread). And definition of the [event](https://developer.mozilla.org/en-US/docs/Web/API/Worker/message_event#event_properties) on the worker side says that `event.ports` is 
> An array of [MessagePort](https://developer.mozilla.org/en-US/docs/Web/API/MessagePort) objects representing the ports associated with the channel the message is being sent through

Maybe this has evolved though, and if you pass a `MessagePort` in the second params, it's automatically pushed as the first element in the `event.ports` on the other side, which would explain why what you have works 🤔

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+  const normalized = username.toLowerCase().trim()
+  const digest = await crypto.subtle.digest('SHA-256', new TextEncoder().encode(normalized))
+  return new Uint8Array(digest)
+}
+
+export type RecoveryKeySignerMessage =
+  | { kind: 'create'; recoveryKey: string; challenge: RecoveryCredentialChallenge }
+  | { kind: 'sign' }
+
+const create = async (recoveryKey: string, challenge: RecoveryCredentialChallenge): Promise<RecoveryKeyAttestation> => {
```
Thibault2ss, line 49: nit: maybe the `create()` function should be called `signRecoveryChallenge` ? (so `messageKind: 'signRecoveryChallenge'` as well) 

also the first argument could be renamed `recoveryKey` -> `password` to be clearer ?

packages/sdk-browser/signers/recoveryKey.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { PasswordWrappedKey } from '../keys/password'
+import { postAsync } from '../utils/messaging'
+
+export type RecoveryCredentialChallenge = Omit<
+  UserRegistrationChallenge,
+  'authenticatorSelection' | 'excludeCredentials' | 'otpUrl'
+> & { challengeIdentifier?: string }
+
+export const randomRecoveryKey = (): string => {
```
Thibault2ss, line 13: nit: we could rename this `randomRecoveryKey` -> `randomRecoveryCode` or `randomRecoveryPassword` otherwise it's a bit confusing feels like this is about the recovery keypair etc...

Currently in the recovery kit PDF, it's called "Recovery Code" I think

PR #78, opened 2024-04-17
drop unused credential transports

djin-dfns:


packages/sdk-react-native/index.ts
```diff
@@ -134,9 +126,9 @@ class iOSPasskeys implements CredentialSigner<Fido2Assertion>, CredentialStore<F
         name: challenge.user.name,
       },
       attestation: challenge.attestation,
-      excludeCredentials: challenge.excludeCredentials.map((cred) => ({
-        id: cred.id,
-        type: cred.type,
+      excludeCredentials: challenge.excludeCredentials.map(({ id, type }) => ({
+        id: b64UrlSafeToStandard(id),
```
ncornwell, line 130: Have we been handling this incorrectly?

packages/sdk-browser/signers/webauthn.ts
```diff
@@ -23,10 +23,9 @@ export class WebAuthnSigner implements CredentialSigner<Fido2Assertion>, Credent
     const response = await navigator.credentials.get({
       publicKey: {
         challenge: Buffer.from(challenge.challenge),
-        allowCredentials: challenge.allowCredentials.webauthn.map(({ id, type, transports }) => ({
+        allowCredentials: challenge.allowCredentials.webauthn.map(({ id, type }) => ({
           id: fromBase64Url(id),
           type,
-          transports: transports ?? [],
```
ncornwell, line 29: Are these helpful at all or are browsers only using the id for lookup?

packages/sdk-react-native/index.ts
djin-dfns: yes, but it's only hit when you create an additional credential. on registration, there's no `excludeCredentials`. I don't think anyone's hit this problem yet since the react native integration is fairly new.

packages/sdk-browser/signers/webauthn.ts
djin-dfns: yeah from our api, the user action challenge we return never has this set. it's always going to be `undefined`. this check here is completely useless.

I'm not totally sure what the behavior would be if we limit the transports. I assume it will disqualify the crendentials that doesn't satisfy the transports requirement. but we don't use it because it's not set by our api.

ncornwell (approved):

PR #86, opened 2024-05-14
Create credential with code example flow

Thibault2ss:
This PR modifies the auth-direct example to rather have existing Dfns users create a new credential, leveraging the [Create Credential With Code](https://docs.dfns.co/d/advanced-topics/authentication/credentials#create-credential-with-code-flow) flow. [This video](https://www.loom.com/share/a8b3cbca4b934f659e2e37e676762b87?sid=6eddb990-3ab9-4b2c-b344-7ccbb7492aaf) showcases the changes to the example.

The PR also:
- generated latest types from the current API to fix the 'public-key' literal string
- refactors the `CredentialStore` class to use the type of the response generated from our auth API endpoints as the input for the `.create(...)` method.


examples/sdk/auth-direct/src/hooks/useAuth.tsx
```diff
@@ -70,8 +75,40 @@ export const AuthProvider = ({ children }: { children: ReactNode }): React.JSX.E
...
+    setLoading(true)
+    setError(undefined)
+
+    try {
+      const challenge = await dfnsApi().auth.createCredentialChallengeWithCode({
+        body: { code, credentialKind: 'Fido2' },
+      })
+
+      if (challenge.kind !== 'Fido2') {
+        throw Error('Not a Fido2 challenge') // check for proper typescript type inferrence
```
adriendfns: is this a todo?

examples/sdk/auth-direct/src/hooks/useAuth.tsx
```diff
@@ -70,8 +75,40 @@ export const AuthProvider = ({ children }: { children: ReactNode }): React.JSX.E
       .finally(() => setLoading(false))
   }
 
+  const createCred = async (code: string): Promise<CreateCredentialWithCodeResponse | undefined> => {
+    setLoading(true)
+    setError(undefined)
+
+    try {
+      const challenge = await dfnsApi().auth.createCredentialChallengeWithCode({
+        body: { code, credentialKind: 'Fido2' },
```
adriendfns, line 84: do we have a `CredentialKind` type?

examples/sdk/auth-direct/src/hooks/useAuth.tsx
Thibault2ss: no it's an indication of why I do this check. I'll update to `meant for proper typescript type inferrence`

examples/sdk/auth-direct/src/hooks/useAuth.tsx
Thibault2ss: in the newly generated SDK, these are string unions rather than enums, so there's no explicit `CredentialKind` enum exported from the `generated/` folder. (even though Intellisense still properly autocompletes & TS checks that this string is correct)

adriendfns (approved):

PR #93, opened 2024-06-10
fix: use the published export/import wasm packages

Thibault2ss:
Update packages `@dfns/sdk-keyimport-utils` & `@dfns/sdk-keyexport-utils` to rather use the external npm libraries `@dfns/dfns-key-import` & `@dfns/dfns-key-export` which contain the rust-generated wasm utils for import/export.

These external libs, used to be built locally and the generated files commited into this repo directly. But now they have their own publish pipeline inside the rust lib they originate from ([trusted-dealer](https://github.com/dfns/trusted-dealer)).


adriendfns (dismissed):

ncornwell (approved):

packages/sdk-keyimport-utils/index.ts
```diff
@@ -1,28 +1,39 @@
...
+  SignersInfo,
+  KeyCurve as KeyCurveInternal,
+  KeyProtocol,
+} from '@dfns/dfns-key-import'
 import { ImportWalletBody, GetWalletResponse } from '@dfns/sdk/types/wallets'
 import { ListSignersResponse } from '@dfns/sdk/types/signers'
 
 type Signer = ListSignersResponse['clusters'][number]['signers'][number]
-type KeyScheme = GetWalletResponse['signingKey']['scheme']
+type KeyCurve = GetWalletResponse['signingKey']['curve']
```
djin-dfns, line 12: is this a breaking change?

packages/sdk-keyimport-utils/index.ts
```diff
@@ -1,28 +1,39 @@
...
+
+const getCurveAndProtocol = (keyCurve: KeyCurve): { curve: KeyCurveInternal; protocol: KeyProtocol } => {
+  switch (keyCurve) {
+    case 'secp256k1':
```
djin-dfns, line 16: not relevant to this PR, but how are we going to support `Schnorr`? it also uses `secp256k1`.

djin-dfns (approved):

packages/sdk-keyimport-utils/index.ts
Thibault2ss: maybe at some point yes 🤷 when we cross that bridge, we’ll need to expose both curve + protocol in the `splitPrivateKeyForSigners()` signature, instead of inferring one as we do now

PR #95, opened 2024-06-11
Ton integration

Torakushi:


examples/libs/ton/basic-tx/main.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    const cell = signingMessageBuilder.endCell()
+
+    const signedCell = await senderWallet.sign(cell)
+    
+    await opened.send(signedCell)
+    console.log('native transfer cell broadcasted')
+
+    // unfortunately, we can't have the txHash without pulling the chain
+    const txHash = await retry(() => tryGetTxByBoc(client, tonWallet.address, signedCell.hash().toString('hex'), 10), {
```
adriendfns, line 84: instead of calling the blockchain directly, can't we just poll DFNS Api to retrieve the hash?

packages/lib-ton/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+    })
+
+    assertSigned(res)
+    if (!res.signedData) {
+      throw new DfnsError(-1, 'signature missing', res)
+    }
+
+    const boc = Cell.fromBoc(hexToBuffer(res.signedData))
+    // This check is already present in the TON sign endpoint
```
adriendfns, line 80: isn't it redundant? I am not sure why we would like to limit that in the client if our API is handling it?

Torakushi: It's just for extra security, but I can remove it if you want

ncornwell (approved):

PR #112, opened 2024-08-21
Kaspa integration

Torakushi:


packages/lib-kaspa/index.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    assertSigned(res)
+    if (!res.signedData) {
+      throw new DfnsError(-1, 'signature missing', res)
+    }
+
+    return this.decodeHexTransaction(res.signedData)
+  }
+
+  public buildTransaction(utxos: any[], to: string, amount: bigint, feePerUtxo: bigint): Transaction {
```
djin-dfns: I rather not be responsible for this. should be similar to bitcoinjs integration. let developers be responsible for constructing the correct transaction however they want, and we will just sign it and return signed transaction back. and also expose a `signHash` method.

ncornwell (approved):

PR #122, opened 2024-10-08
feat: publish 0.5.7

ncornwell:


packages/lib-kaspa/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dfns/lib-kaspa",
-  "version": "0.4.3",
+  "version": "0.5.7",
```
wendymock, line 3: big jump here

ncornwell: I don't think it's been published before 😄

wendymock (approved):

PR #128, opened 2024-10-31
DFNS-2367 Update SDK and Add taproot example

adriendfns:


packages/lib-bitcoinjs/index.ts
```diff
@@ -81,6 +119,11 @@ export class DfnsWallet implements SignerAsync {
 
     return Psbt.fromHex(res.signedData.replace(/^0x/, ''))
   }
+
+  public tweakTaprootMerkleRoot(taprootMerkleRoot: Buffer) {
+    this.taprootMerkleRoot = taprootMerkleRoot
+    this.publicKey = getTapInternalKey(this.rootPublicKey, this.taprootMerkleRoot, true)
```
Thibault2ss: nit: instead of setting `this.publicKey` in different places, we could implement it as a getter maybe ? which depends on `this.rootPublicKey` / `this.scheme` / `this. taprootMerkleRoot`

adriendfns: unfortunately we have `export class DfnsWallet implements SignerAsync`
this object is then passed to the bitcoin library, it needs to call `publicKey`.

An alternative would be to clone when tweaking, so it would be something like:

```
const wallet = = await initDfnsWallet(process.env.BITCOIN_WALLET_ID!)
const wallet2 = wallet.tweak(taprootMerkleRoot)
```

djin-dfns: hmm, I don't like the mutating state inside the wallet, but also doesn't make sense to model the tweak as a separate wallet, the underlying wallet is the same. you are just tweaking the signing. the way I would implement this is leave the `DfnsWallet` to only sign with the root key, or tweaked root key for schnorr, introduce a new function `tweakSigner` to return an tweaked signer as an anonymous class for signing with a different merkle root.

```
class DfnsWallet {
 ...

  public tweakSigner(taprootMerkleRoot: Buffer): SignerAsync {
    const tweakedKey = tweakKey(this.publicKey, taprootMerkleRoot)

    return new class implements SignerAsync {
      get publicKey() {
        return tweakedKey
      }

      async signSchnorr(hash: Buffer): Promise<Buffer> {
        // sign with merkle root
      }
    }
  }
}


const tweakedSigner = wallet.tweakSigner(taprootMerkleRoot)
await psbt.signInputAsync(1, tweakedSigner)
```

packages/lib-bitcoinjs/index.ts
```diff
@@ -23,13 +28,38 @@ const assertSigned = (res: GenerateSignatureResponse) => {
...
+  const tweakedPublicKey = ecc.xOnlyPointAddTweak(schnorrPublicKey, tweak)
+  if (parityByte) {
+    return Buffer.concat([
+      Uint8Array.from([tweakedPublicKey!.parity === 0 ? 0x02 : 0x03]),
+      tweakedPublicKey!.xOnlyPubkey,
+    ])
+  } else {
+    return Buffer.concat([tweakedPublicKey!.xOnlyPubkey])
+  }
+}
```
djin-dfns: this function is unnecessary, `bitcoinjs` already has this utility function.

```
import { tweakKey } from 'bitcoinjs-lib/src/payments/bip341'
```

instead, let's add a new getter/property to `DfnsWallet`, then it's easy to tweak the public key `tweakKey(wallet.xOnlyPublicKey, taprootMerkleRoot)`

```
public get xOnlyPublicKey() {
  return this.publicKey.subarray(1)
}
```

adriendfns: If I understand your message correctly, you want our customers to derive the public key themselves. I don't think we can take this path:

- Because we are implementing `SignerAsync`, we need the tweaked signer to have a tweaked public key, so we need to compute the new public key anyway
- The library does not allow the hash to be empty: https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/cjs/payments/bip341.cjs#L158, unfortunately, for the default address (empty merkleroot) we cannot use it

packages/lib-bitcoinjs/index.ts
```diff
@@ -81,6 +117,10 @@ export class DfnsWallet implements SignerAsync {
 
     return Psbt.fromHex(res.signedData.replace(/^0x/, ''))
   }
+
+  public tweakSigner(taprootMerkleRoot: Buffer): SignerAsync {
+    return new DfnsWallet(this.metadata, { ...this.options, taprootMerkleRoot })
```
djin-dfns: I prefer to return an anonymous class `SignerAsync` that's locked down to only do Schnorr signing and nothing else. see my previous comment.

conceptually, tweaked signer does not map to a Dfns wallet. for instance, `signPsbt` function doesn't make sense to have on the tweaked signer. let's make the design as tight as possible to avoid people do things that ends up not what they expected. tweaked signer should implement the bare minimum that's needed, `publicKey` + `signSchnorr`. also regular `sign` should throw error.

packages/lib-bitcoinjs/index.ts
djin-dfns: it [does support](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/cjs/payments/bip341.cjs#L146) empty/null/undefined merkleroot, here's the test

```
const key = ECPair.fromPrivateKey(
  Buffer.from(
    "bff66a7299bd8bcd1e9ddbe51e5aadf4b51a61da474b55c23dcd1d87b2d04264",
    "hex"
  )
);

console.log(tweakKey(key.publicKey.subarray(1), undefined));

console.log(
  tweakKey(
    key.publicKey.subarray(1),
    Buffer.from(
      "0000000000000000000000000000000000000000000000000000000000000000",
      "hex"
    )
  )
);
```

> If I understand your message correctly, you want our customers to derive the public key themselves.

scratch the idea of `xOnlyPubKey` then, I didn't consider the root wallet also needs to return the tweaked public key, we can change the publicKey to conditionally return either the original one for ecdsa or tweaked one for schnorr, we don't ever need to expose the original if schnorr right?

```
private _publicKey = Buffer.from(publicKey, 'hex)

public get publicKey() {
  if (Schnorr) {
    return tweakKey(this._publicKey.subarray(1), undefined)
  } else {
    return this._publicKey
  }
}
```

djin-dfns: > If I understand your message correctly, you want our customers to derive the public key themselves.

I guess it's still unclear to me, should `publicKey` be the original key or the tweak key for a Dfns wallet. does the property get used at all when `signInputAsync(signer)` is called?

> Because we are implementing SignerAsync, we need the tweaked signer to have a tweaked public key, so we need to compute the new public key anyway

justification in my other comment, tweaked signer should be completely independent from the `DfnsWallet` class, because I find mixing them into one class is confusing. the behavior of the class becomes hard to reason through. also we can compute the key by using the bitcoin js helper. my original comment is we don't need to duplicate the logic.

> The library does not allow the hash to be empty

it [does support](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/cjs/payments/bip341.cjs#L146) empty/null/undefined merkleroot, here's the test

```
const key = ECPair.fromPrivateKey(
  Buffer.from(
    "bff66a7299bd8bcd1e9ddbe51e5aadf4b51a61da474b55c23dcd1d87b2d04264",
    "hex"
  )
);

console.log(tweakKey(key.publicKey.subarray(1), undefined));

console.log(
  tweakKey(
    key.publicKey.subarray(1),
    Buffer.from(
      "0000000000000000000000000000000000000000000000000000000000000000",
      "hex"
    )
  )
);
```

examples/libs/bitcoinjs/taproot/main.ts
```diff
@@ -0,0 +1,266 @@
...
+  });
+
+  await psbt.signInputAsync(0, wallet);
+  psbt.finalizeAllInputs();
+
+  const tx = psbt.extractTransaction();
+  console.log(`Broadcasting Transaction Hex: ${tx.toHex()}`);
+  const txid = await broadcast(tx.toHex());
+  console.log(`Success! Txid is ${txid}`);
+}
```
djin-dfns: I like to keep example focused. this first function is just a simple transfer with a taproot key. basically the same as the `basic-tx` example, not very interesting. we should just drop it.

if you think there's value keeping it, then make two separate examples.

examples/libs/bitcoinjs/taproot/main.ts
```diff
@@ -0,0 +1,266 @@
...
+
+
+async function start_taptree(wallet: DfnsWallet) {
```
djin-dfns: nit: function and variable names are not following ts naming convention.

examples/libs/bitcoinjs/taproot/main.ts
```diff
@@ -0,0 +1,266 @@
...
+  utxos = await waitUntilUTXO(script_addr)
+  console.log(`Trying the P2PK spend path with UTXO ${utxos[0].txid}:${utxos[0].vout}`);
+
+  const key_spend_psbt = new Psbt({ network });
+  key_spend_psbt.addInput({
+    hash: utxos[0].txid,
+    index: utxos[0].vout,
+    witnessUtxo: { value: utxos[0].value, script: script_p2tr.output! },
+    tapInternalKey: toXOnly(wallet.rootPublicKey),
+    tapMerkleRoot: script_p2tr.hash
```
djin-dfns: so we place the `taprootMerkleRoot` in the psbt input, doesn't this mean we should be able to ship the whole psbt server side to sign?

packages/lib-bitcoinjs/index.ts
adriendfns: > I guess it's still unclear to me, should publicKey be the original key or the tweak key for a Dfns wallet

When the signer is tweaked, it needs to return the tweaked public key, `signInputAsync` will call `signer.publicKey` we have no choice but to modify it

>  tweaked signer should be completely independent from the DfnsWallet class

I don't fully agree, maybe we can huddle about it

> it [does support](https://github.com/bitcoinjs/bitcoinjs-lib/blob/master/src/cjs/payments/bip341.cjs#L146) empty/null/undefined merkleroot, here's the test

my bad, it is now fixed

examples/libs/bitcoinjs/taproot/main.ts
adriendfns: Yes I think we could extract it, however we won't be able to do just:

```
 const psbt = Psbt.fromBuffer(hexToBuffer(unsignedData))
 await psbt.signAllInputsAsync(signer, [Transaction.SIGHASH_ALL])
```

because in the `SignerAsync` callbacks, `signSchnorr?(hash: Buffer)` does not have any hint on what is the `taprootMerkleRoot`

we would have to iterate through the inputs and sign one by one.
I guess it is doable, do you think we should update the bik code?

packages/lib-bitcoinjs/index.ts
```diff
@@ -1,13 +1,19 @@
 import { DfnsApiClient, DfnsError } from '@dfns/sdk'
 import { GetWalletResponse, GenerateSignatureResponse } from '@dfns/sdk/types/wallets'
 import { Network, Psbt, SignerAsync } from 'bitcoinjs-lib'
+import { tweakKey } from 'bitcoinjs-lib/src/payments/bip341'
+import * as ecc from 'tiny-secp256k1'
+import { initEccLib } from "bitcoinjs-lib";
+
+initEccLib(ecc)
```
djin-dfns: just occurred to me, we should leave this to the sdk user to do the init, just like bitcoinjs. we shouldn't assume people will use schnorr, and they can add this init line if they want to use schnorr, and we don't need to take a dep on tiny secp lib.

packages/lib-bitcoinjs/index.ts
```diff
@@ -52,9 +77,17 @@ export class DfnsWallet implements SignerAsync {
...
+
+  public async signSchnorr(hash: Buffer): Promise<Buffer> {
+    return this._sign(hash, Buffer.alloc(0))
+  }
+
+  private async _sign(hash: Buffer, taprootMerkleRoot?: Buffer): Promise<Buffer> {
     const res = await this.dfnsClient.wallets.generateSignature({
       walletId: this.metadata.id,
-      body: { kind: 'Hash', hash: `0x${hash.toString('hex')}` },
+      body: { kind: 'Hash', hash: `0x${hash.toString('hex')}`, taprootMerkleRoot: taprootMerkleRoot && taprootMerkleRoot.toString('hex') },
```
djin-dfns: nit: add a `0x` prefix to the merkle root hex. our api is lenient on the hex formatting, but our own lib should try to be formatting consistent.

```suggestion
      body: { kind: 'Hash', hash: `0x${hash.toString('hex')}`, taprootMerkleRoot: taprootMerkleRoot && `0x${taprootMerkleRoot.toString('hex')}` },
```

packages/lib-bitcoinjs/index.ts
```diff
@@ -81,6 +114,25 @@ export class DfnsWallet implements SignerAsync {
 
     return Psbt.fromHex(res.signedData.replace(/^0x/, ''))
   }
+
+  public tweakSigner(taprootMerkleRoot: Buffer): SignerAsync {
+    const publicKey = getTweakedPublicKey(this.internalPubkey, taprootMerkleRoot)
```
djin-dfns: I think should have a check here to only allow tweaking for schnorr.

packages/lib-bitcoinjs/index.ts
```diff
@@ -81,6 +114,25 @@ export class DfnsWallet implements SignerAsync {
...
+    const wallet = this
+
+    return new (class implements SignerAsync {
+      get publicKey() {
+        return publicKey
+      }
+
+      async sign(_hash: Buffer): Promise<Buffer> {
+        throw 'Expecting to only do schnorr signatures'
```
djin-dfns: let's throw a DfnsError.

packages/lib-bitcoinjs/index.ts
```diff
@@ -1,13 +1,15 @@
 import { DfnsApiClient, DfnsError } from '@dfns/sdk'
 import { GetWalletResponse, GenerateSignatureResponse } from '@dfns/sdk/types/wallets'
 import { Network, Psbt, SignerAsync } from 'bitcoinjs-lib'
+import { tweakKey } from 'bitcoinjs-lib/src/payments/bip341'
 
 export type DfnsWalletOptions = {
   walletId: string
   dfnsClient: DfnsApiClient
+  taprootMerkleRoot?: Buffer
```
djin-dfns: don't need this anymore.

packages/lib-bitcoinjs/index.ts
```diff
@@ -23,13 +25,32 @@ const assertSigned = (res: GenerateSignatureResponse) => {
...
+  public readonly scheme: GetWalletResponse['signingKey']['scheme']
+  public publicKey: Buffer
+  public internalPubkey: Buffer
 
   private constructor(private readonly metadata: WalletMetadata, options: DfnsWalletOptions) {
     this.dfnsClient = options.dfnsClient
+    this.scheme = metadata.signingKey.scheme
     this.publicKey = Buffer.from(metadata.signingKey.publicKey, 'hex')
+
+    // If type is Schnorr, we set the "default" value for the taproot merkle root
```
djin-dfns: nit: comment outdated

packages/lib-bitcoinjs/index.ts
```diff
@@ -23,13 +25,32 @@ const assertSigned = (res: GenerateSignatureResponse) => {
...
+  const tweakedKey = tweakKey(xOnlyPublicKey, taprootMerkleRoot)
+  return Buffer.concat([Uint8Array.from([tweakedKey!.parity + 2]), tweakedKey!.x])
+}
+
 export class DfnsWallet implements SignerAsync {
   private readonly dfnsClient: DfnsApiClient
-  public readonly publicKey: Buffer
+  public readonly scheme: GetWalletResponse['signingKey']['scheme']
+  public publicKey: Buffer
+  public internalPubkey: Buffer
```
djin-dfns: it's nullable since we don't set it if not schnorr.

```suggestion
  public internalPubkey?: Buffer
```

djin-dfns (approved):

PR #159, opened 2025-03-17
feat(iota): add iota sdk

Torakushi:


Torakushi:
@SocketSecurity ignore npm/@gql.tada/cli-utils@1.6.3

packages/lib-iota/index.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export type SignedTransaction = {
+  // transaction block as base64 encoded bcs
+  transactionBlock: string
+  // base64 encoded signatures. Usually, there will be only one signature
+  signature: string[]
+} 
+
+export class DfnsWallet {
```
djin-dfns: we should implement the [signer interface](https://github.com/iotaledger/iota/blob/develop/sdk/typescript/src/cryptography/keypair.ts#L33), mainly the `signTransaction` function, we can throw not supported error for the other functions for now. advantage is once we satisfy that base class, then our signer can be passed to [`client.signAndExecuteTransaction`](https://github.com/iotaledger/iota/blob/develop/sdk/typescript/src/client/client.ts#L430), which can trim down the examples a bit, generally make the usage slightly friendlier I think.

examples/libs/iota/basic-tx/main.ts
```diff
@@ -0,0 +1,59 @@
...
+  tx.transferObjects([coin], recipient)
+  tx.setSender(wallet.address)
+
+  const txBytes = await tx.build({ client })
+
+  const signedTx = await wallet.sign(txBytes)
+
+  console.log(`transaction signed`)
+
+  const res = await client.executeTransactionBlock(signedTx)
```
djin-dfns: after implementing as `Signer`, this can be trimmed to

```suggestion
  const tx = new Transaction()
  const [coin] = tx.splitCoins(tx.gas, [amount])
  tx.transferObjects([coin], recipient)
  const res = await client.signAndExecuteTransaction(tx, wallet)
```

djin-dfns (approved):

PR #166, opened 2025-03-28
Add viem multicall example

adriendfns:


Thibault2ss (dismissed):

examples/libs/viem/multicall/main.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  return DfnsWallet.init({ walletId, dfnsClient })
+}
+
+const main = async () => {
+  // The sponsor wallet will pay for the gas
+  const ethereumWallet = await initDfnsWallet(process.env.ETHEREUM_WALLET_ID!);
+
+  const ethereumWalletClient = createWalletClient({
+    account: toAccount(ethereumWallet as any),
```
djin-dfns: is this cast necessary? feels like if we need to cast here, then our integration is off on the typing somewhere.

examples/libs/viem/multicall/multicallAbi.ts
```diff
@@ -0,0 +1,55 @@
+export const multicallAbi = [
```
djin-dfns, line 1: nit: if we can inline a [human readable version](https://www.multicall3.com/abi#viem) of the slim abi in main.ts, I find that's more clear for example demonstration purpose.

adriendfns: I am not sure how to parse this, if I use it directly I get:

```
/dfns-sdk-ts/node_modules/viem/utils/abi/getAbiItem.ts:90
  const abiItems = (abi as Abi).filter((abiItem) => {

TypeError: Cannot use 'in' operator to search for 'name' in struct Call { address target; bytes callData; }
  ```

djin-dfns: ok, we can figure it out later.

djin-dfns (approved):

PR #167, opened 2025-04-04
feat: bump version 0.6.7

ncornwell:


packages/sdk/generated/wallets/types.ts
```diff
@@ -154,6 +154,7 @@ export type ExportWalletBody = {
         protocol: "CGGMP21" | "FROST" | "FROST_BITCOIN" | "KU23";
         curve: "ed25519" | "secp256k1" | "stark";
     }[];
+    delete?: boolean | undefined;
```
Thibault2ss, line 157: Should we change flag name before we release that to SDK ? Otherwise changing later will be breaking, and SDK is probably where people discover our API updates first, rather than docs. (cf [this comment](https://github.com/dfnsco/dfns/pull/1945#discussion_r2026567104))

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.6.6 (2024-04-03)
```
Thibault2ss: 0.6.6 -> 0.6.7

ncornwell: Fixed

djin-dfns (approved):

PR #180, opened 2025-06-10
feat: deprecate applications

Thibault2ss:
Deprecating Dfns "Applications" in the SDK.

Read [this article](https://docs.dfns.co/d/guides/applications-depreaction) for more context


packages/sdk/utils/fetch.ts
```diff
@@ -75,25 +75,20 @@ export const catchPolicyPending = <T>(fetch: Fetch<T>): Fetch<T> => {
...
 
-    const dfnsAppSecret: Record<string, string> = appSecret
-      ? {
-          'x-dfns-appsecret': appSecret,
-        }
-      : {}
-
     options.headers = {
-      'x-dfns-appid': appId,
-      'x-dfns-nonce': generateNonce(),
```
adriendfns, line 94: nonce is not a thing anymore?

adriendfns (dismissed):

packages/sdk/utils/fetch.ts
Thibault2ss: it was only a thing in the context of server-signed apps, otherwise, the backend was not really using it. And since for all non-GET requests, we have a user action token with unique ID, which can only be used once (basically acting as a nonce), we don't need this one anymore

adriendfns (approved):

PR #197, opened 2025-08-04
feat(near): sdk integration

Torakushi:


adriendfns (dismissed):

Torakushi:
@SocketSecurity ignore-all.

packages/lib-near/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      body: {
+        kind: 'Transaction',
+        transaction: '0x' + Buffer.from(transaction.encode()).toString('hex'),
+      },
+    })
+
+    assertSigned(res)
+    if (!res.signature?.encoded) {
+      throw new DfnsError(-1, 'signature missing', res)
+    }
```
djin-dfns: nit, doesn't look like we are using this, so this check seems unnecessary. though it should not fail anyway if the signed data is present.

packages/lib-near/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@dfns/lib-near",
+  "version": "0.7.5",
+  "peerDependencies": {
+    "near-api-js": "6.2.4"
```
djin-dfns: make it a bit more flexible

```suggestion
    "near-api-js": "^6.0.0"
```

djin-dfns (approved):

PR #204, opened 2025-09-03
abortable webauthn

djin-dfns:


djin-dfns:
not supported by any authenticators. even though theoretically this should work. close for now.

PR #211, opened 2025-09-11
feat: udpate nx

ncornwell:


package.json
```diff
@@ -32,23 +32,23 @@
     "@solana/spl-token": "0.4.12",
     "@stablelib/blake2b": "1.0.1",
     "@tronweb3/tronwallet-abstract-adapter": "1.1.6",
-    "buffer": "6.0.3",
+    "buffer": "^6.0.3",
```
Thibault2ss: we wanna fix that version ?

packages/sdk-browser/package.json
```diff
@@ -1,4 +1,7 @@
 {
   "name": "@dfns/sdk-browser",
-  "version": "0.7.10"
+  "version": "0.7.10",
+  "dependencies": {
+    "buffer": "^6.0.3"
```
Thibault2ss: - fix version ?
- If that work for now, fine. Though ideally we figure out why we have to add that manually given `buffer` is not automatically detected & added by `@nx/js:tsc` in the generated package.json (in `dist/`)
- We should add this package to all packages which explicitly import it, so both `@dfns/sdk` and `@dfns/sdk-browser` I believe

Thibault2ss (commented):
Today all other packages, when built, have some dependencies in their `package.json`. We need to compare all of them so see if nx properly added those as well

examples/libs/ethersjs/v6/base-deposit-eth/main.ts
```diff
@@ -27,25 +27,38 @@ const initDfnsWallet = async (walletId: string) => {
...
+  })
+  const usdc = '0x036cbd53842c5426634e7929541ec2318f3dcf7e'
 
-  const l1Address = await ethWallet.getAddress()
-  console.log(`Ethereum L1 sending address ${l1Address}`)
+  const erc20 = new Interface(['function transfer(address to, uint amount)'])
+  const data = erc20.encodeFunctionData('transfer', ['0x416a2003ba6e8c2ee25816a8cbd09dca187049b3', 1])
 
-  const l2Address = await baseWallet.getAddress()
-  console.log(`Base L2 receiving address ${l2Address}`)
```
Thibault2ss, line 39: Is this example change supposed to be included ?

ncornwell: No, I'm going to close this pr as the general update is already ahead

PR #231, opened 2025-09-17
feat(hedera): add hedera

Torakushi:


packages/lib-hedera/index.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  public getMirrorNetwork(): string[] {
+    return this.provider.getMirrorNetwork()
+  }
+
+  public getAccountKey(): PublicKey {
+    return this.publicKey
+  }
+
+  public getAccountBalance(): Promise<AccountBalance> {
```
Torakushi, line 160: We could have left the error ‘throw not implemented’. But since it was easy to implement... let's do it so that the user can eventually use it for complex operations.

Torakushi:
@SocketSecurity ignore-all

ncornwell (approved):

PR #232, opened 2025-09-17
fix: upgrade meshsdk

ncornwell:


package.json
```diff
@@ -58,7 +58,7 @@
     "@cosmjs/crypto": "^0.32.0",
     "@cosmjs/encoding": "^0.32.0",
     "@cosmjs/proto-signing": "^0.32.0",
-    "@meshsdk/core": "^1.0.0",
+    "@meshsdk/core": "^1.8.14",
```
djin-dfns, line 61: can we make this

```suggestion
    "@meshsdk/core": ">=1.8.0",
```

we are using peer deps here to not hard fixed to a version. we've had feedback that when hardfixed to a version, people had problem integrating our sdk with other toolings.

djin-dfns: actually I wonder if you can just do a `npm update` to upgrade all the loose deps and not need to change the pinned versions at all.

PR #235, opened 2025-09-18
feat(kadena): add kadena sdk

Torakushi:


examples/libs/kadena/basic-tx/main.ts
```diff
@@ -0,0 +1,84 @@
...
+  const hoppAmount = "1"
+  const pactAmount = { decimal: toKda(hoppAmount).toString() }
+  console.log(`transferring ${hoppAmount} Hopp from wallet ${senderWallet.address} to ${recipient}`)
+  console.log(`network: ${senderWallet.networkId}, chain: ${senderWallet.chainId}`)
+
+  // you can change the token to transfer
+  const contract = 'coin'
+
+  console.log('creating transfer command...')
+  const transferCommand = Pact.builder
```
Torakushi, line 56: Even if it looks complex, we should let the client build the PactCommand IMO. The goal of the SDK is to ease signature, not to provide a `COnstruct` EP.

ncornwell (dismissed):

ncornwell (approved):

PR #282, opened 2025-11-26
feat: allow for async signing on solana sdk

ncornwell:


packages/lib-solana/index.ts
```diff
@@ -80,11 +59,26 @@ export class DfnsWallet {
         },
       })
 
-      assertSigned(res)
-      if (!res.signedData) {
+      let responseStatus = res.status
+      let signedData = res.signedData
+      while (!signedData && responseStatus != 'Failed') {
```
Torakushi, line 64: Maybe we can extract this in a different function so we don't have this copy-paste in 3 different places?

PR #332, opened 2026-02-26
feat(starknet): add starknet integration

pierre-dfns:


pierre-dfns:
@SocketSecurity ignore npm/cardinal@2.1.1

packages/lib-starknet/index.ts
```diff
@@ -0,0 +1,218 @@
...
+  public get address(): string {
+    return this.walletAddress
+  }
+
+  public async getPubKey(): Promise<string> {
+    return this.publicKeyHex
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public async signMessage(_typedData: TypedData, _accountAddress: string):
```
djin-dfns, line 94: this is snip12, we do support this.

pierre-dfns: that's true... I was too focused on signing transaction :D 
thanks!

PR #355, opened 2026-03-26
ci: semantic release + publish from CI

xavier-dfns:
## Summary

- Add automated semantic versioning and npm publishing via CI, replacing the manual bin/version_all.sh and bin/publish.sh scripts.
- Add release.yaml workflow: runs after a successful build on m, uses semantic-release to determine the next version, bumps all workspace packages, commits the version bump, and creates a GitHub release
- Add publish.yaml workflow: triggered by a GitHub release, builds and publishes all packages to npm. Migrate to OIDC / npm trusted publisher instead.
- Pin all GitHub Actions to commit SHAs across build.yaml and docs.yaml for supply-chain security
- Add .releaserc.yml for semantic-release configuration


james-dfns (approved):

djin-dfns:
if I'm not mistaken, every merge to m will cause a release to be pushed to npm? if that's the case, then that's not what we want. because this could cause unreleased changes to go out in the sdk before our server release is out.

xavier-dfns:
@djin-dfns I could change the workflow to trigger on push on a different branch (like `release` ?). Would that be satisfying ?

PR #359, opened 2026-03-27
ci: add protected branch bypass

xavier-dfns:
Resolve issue whereas github bot cannot push commit to protected branch due to ruleset we have on `m`.
Bypass can't be created for github action token but for github app it can.


james-dfns (approved):
\+ one less dependency 🔥

PR #718, opened 2026-09-09
fix(lib-concordium): pin @concordium/web-sdk peer to 12.0.2

jonas-dfns:
The `@concordium/web-sdk` peer in `packages/lib-concordium` was pinned to the exact prerelease `12.0.0-alpha.3`. Any workspace bump of `@concordium/web-sdk` past that version therefore failed `npm install` with `ERESOLVE` — which is what blocked the grouped Dependabot PR #713 (it tries to move the root to `12.0.2`).

## Change
- `packages/lib-concordium`: widen the peer from `12.0.0-alpha.3` → `>=12.0.0-alpha.3 <13.0.0` (still accepts the old prerelease, now also the stable 12.x line).
- root `package.json`: adopt `@concordium/web-sdk@12.0.2` in the dev harness (which resolves against the workspace `lib-concordium`), regenerating the lockfile.

## Verification
`npm ci` + `npm run cb:all` (the `build`/`compile` job) on Node 24 in a clean container: install resolves with no `ERESOLVE`, and all 33 workspace projects compile — `lib-concordium` builds unchanged against `12.0.2` (it only uses the stable `AccountAddress` / `Transaction` APIs).

## Scope
web-sdk only. The example manifests still consume the *published* `@dfns/lib-concordium` (which peers the old `alpha.3` until this ships in the next release), so their web-sdk pins are intentionally left at `alpha.3` for now. Once a release publishes the widened peer, #713 can rebase (its `@concordium/web-sdk` bump becomes a no-op) and merge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "@aptos-labs/ts-sdk": "^5.0.0",
         "@aws-sdk/client-kms": "^3.1027.0",
         "@concordium/id-app-sdk": "0.1.4",
-        "@concordium/web-sdk": "12.0.0-alpha.3",
+        "@concordium/web-sdk": "12.0.2",
         "@dfns/dfns-key-export-bundler": "0.4.0",
         "@dfns/dfns-key-export-nodejs": "0.4.0",
         "@dfns/dfns-key-import-bundler": "0.4.0",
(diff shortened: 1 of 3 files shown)


pierre-dfns:
@SocketSecurity ignore npm/@concordium/web-sdk@12.0.2

pierre-dfns (approved):

